PR #19, opened 2022-10-05
Model and import area data

struan:


.env-example
```diff
@@ -1,2 +1,4 @@
 SECRET_KEY="this_IS_secr3t"
 DEVENV="docker"
+MAPIT_URL="https://mapit.mysociety.org/"
```
zarino, line 3: Should we also add the `MAPIT_URL` setting to the `web` container’s environment in `docker-compose.yml`?

.env-example
```diff
@@ -1,2 +1,4 @@
...
+MAPIT_API_KEY=""
```
zarino, line 4: I wonder what the best way to handles this is, for Docker users. Perhaps `README.md` should be updated, to include an instruction about creating a `.env` file with your MapIt API key in it, before running `docker-compose up`?

hub/management/commands/import_areas.py
```diff
@@ -0,0 +1,39 @@
+import json
+from django.core.management.base import BaseCommand
+
+from time import sleep
+from utils import mapit
+from hub.models import Area
+
+
+class Command(BaseCommand):
+    help = "Import basic area information from MaPit"
```
zarino, line 10: 10 bonus points for the archaic capitalisation of MapIt. :+1:

I wonder whether `README.md` should explain that you’ll need to run `./manage.py import_areas` after you’ve started the docker container?

zarino: Might also want to let them know it’ll take at least 10 minutes to run ;-)

hub/static/js/all.js
```diff
@@ -153,18 +155,22 @@ var setUpMap = function() {
...
-                mouseover: highlightFeature,
-                mouseout: unhighlightFeature,
-                click: function(){
-                    window.location.href = '/area'
-                }
-            })
-        }
-    }).addTo(map)
+    $.getJSON("/filter_areas/", function(data) {
+        console.log(data);
```
zarino, line 159: Can see why this was left in, but we should probably remove it before we forget!

zarino (approved):
This all worked for me, once I’d rebuilt the `web` image (to include `requests`) and created an `.env` file with my `MAPIT_API_KEY`.

I’ve left a few comments below, thinking primarily about how we should document setting this stuff up in Docker.

alexander-griffen (approved):
Runs on my machine well!

PR #20, opened 2022-10-19
Add script for creating a Django superuser

zarino:
This adds a script that creates a superuser, using details from the `DJANGO_SUPERUSER_*` environment variables (defaults to the values in docker-compose.yml).

@struan not sure whether we want to put this into `script/update`, or whether to leave it as a manual step? I figure _anyone_ running this code locally will want to set up a superuser… but really we’d only want it to be run when the database is _first_ set up (the user might delete the default superuser account, and not want it auto-recreated by `script/update`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,3 +35,10 @@ management commands in the docker web container:
 * `./manage.py import_mps`
 
 You can then view it at (http://localhost:8000/)
+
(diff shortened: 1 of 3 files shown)


struan (approved):
er, I think probably leave it out of update as it's only really a one off script.

PR #21, opened 2022-10-19
Import basic MP data

struan:
Imports and display some basic MP data from wikidata.


struan:
I was going to do the photos separately.

And the election data is next on the list so it'll stop being static shortly

alexander-griffen (approved):

Issue #8, opened 2022-09-27
Set up staging server


struan:
Initial requirements:

- Django
- Postgres

PR #22, opened 2022-10-20
Add initial example files needed by our deployment automation

sagepe:
I think this is all we'll need at first.

See: https://github.com/mysociety/local-intelligence-hub/issues/8


struan (approved):
looks good.

PR #39, opened 2022-10-27
allow update of area geometries and lessen complexity

struan:
previous complexity was a bit much which resulted in a large download so increase the mapit simplify number.

Also update the import command to always update the geometry and get rid of the sleep.

### DIFF
--- a/hub/management/commands/import_areas.py
+++ b/hub/management/commands/import_areas.py
@@ -28,12 +28,12 @@ class Command(BaseCommand):
             except:
                 geom = None
 
-            a = Area.objects.get_or_create(
+            a, created = Area.objects.get_or_create(
                 mapit_id=area["id"],
                 gss=area["codes"]["gss"],
                 name=area["name"],
                 area_type="WMC",
-                geometry=geom,
             )
 
-            sleep(1)
+            a.geometry = geom
+            a.save()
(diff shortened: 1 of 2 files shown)


struan:
This is a compromise between size of the geometry data downloaded and not having the map data look too janky.

zarino (commented):
5 MB still seems pretty big, especially when compared to the `constituencies.esm.js` we handily happen to have lying around, which was only 800 MB for (I think?) effectively the same data.


## /filter_areas

```
… "Polygon", "coordinates": [[[-3.682993252029648, 52.986874821956334], …
```

Screenshot 2022-10-27 at 16-01-37 Explore

## constituencies.esm.js

```
… "Polygon","coordinates":[[[-0.8048,51.2451],[-0.7761,51.242], …
```

Screenshot 2022-10-27 at 16-01-44 National data explorer Local Intelligence Hub

Admittedly, the resolution’s lower in `constituencies.esm.js`, but does that level of accuracy really matter for a _national_ view?

zarino:
Ah, I guess when I used https://mapshaper.org to simplify the geometries in the prototype, it must have done something clever to avoid those gaps/overlaps.

alexander-griffen (approved):

PR #40, opened 2022-10-27
Add cache busting static file generation

struan:
Use manifest static file storage so changes to CSS/JS don't have to wait for cache expiry.


zarino (changes requested):
I can’t comment on the merits of `ManifestStaticFileStorage`, but I _do_ think you’ll want to keep `leaflet-1.8.0.min.js` around as, even though we don’t load it directly, it’s imported by `leaflet-1.8.0.esm.js` which we _do_ load. Unless I’ve missed something?

zarino (approved):
Gah! I am an idiot! Go ahead 👍

PR #46, opened 2022-11-01
Add constituency age data

### LINKED ISSUE #33: Import constituency age data
https://commonslibrary.parliament.uk/constituency-statistics-population-by-age/

We’re doing this now because it’s a good example of a dataset that has multiple related datapoints per constituency (eg: we’ll want to display the proportion of citizens in _each_ of the age bands, in the _right order_, on a single page).

struan:
Fixes #33

Screenshot 2022-11-11 at 14 58 45


zarino:
FYI @struan I’ve rebased this branch onto a more recent `main`, featuring the new `area.html` page layout.

alexander-griffen (approved):

zarino (approved):
Looks good to me. I’ve added a commit that renders the data as a `<table>` element, and then uses JavaScript to convert it into a bar chart.

PR #47, opened 2022-11-02
postcode/mp/area lookup on front page

struan:
Place/MP/Postcode search on homepage now works. With a disambiguation page when you enter a search term that matches multiple places or MPs:

Screenshot 2022-11-03 at 16-17-36 Screenshot

### DIFF
--- a/hub/models.py
+++ b/hub/models.py
@@ -24,6 +24,9 @@ class Area(models.Model):
     area_type = models.CharField(max_length=20)
     geometry = models.TextField(blank=True, null=True)
 
+    def get_absolute_url(self):
+        return f"/area/{self.area_type}/{self.name}"
+
 
 class Person(models.Model):
     person_type = models.CharField(max_length=10)
@@ -35,6 +38,10 @@ class Person(models.Model):
 
     def __str__(self):
         return self.name
+
+    def get_absolute_url(self):
+        area = self.area
+        return f"/area/{area.area_type}/{area.name}"
 
     class Meta:
         unique_together = ("external_id", "id_type")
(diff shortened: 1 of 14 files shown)


alexander-griffen (approved):
Looks good and works!

Only question is (and I'm not sure if this is a feature or a bug), when I'm using firefox the 'Use my location' doesn't show up (and when inspecting it, 
```
<a class="js-geolocate d-none d-flex align-items-center btn btn-link px-0" href="{% url 'area_search' %}">
                    {% include 'hub/includes/icons/geolocate.html' with classes="me-2" %}
                    Use my current location
</a>
```
is greyed out), however it does show up - and work - on Safari.

zarino:
@alexander-griffen interesting! So once the page has finished loading, the link still has its `d-none` class? If so, then `    if( 'geolocation' in navigator )` must be returning false on your machine. I would have guessed it might be because you’re accessing the site under `http://` rather than `https://` (geolocation only works under `https://`) – but since it works in Safari, and Safari (version 10+) requires https for geolocation too, I’m not sure what might be going on 🤔

alexander-griffen:
@zarino 

Yep, it still has that class. I had a look through my Firefox settings to see if I'd unwittingly set something to prevent geolocation things from showing up, but I can't find any. I did find a https-only setting, which I turned on, and still can't get it to show up.

zarino:
In fact, I’ve just tried it myself, and realised I was getting confused with the Jekyll prototypes, and I’ve never actually tried to access this version of the site over https://, I’ve just been using http:// and of course, the geolocation button remains hidden. And from the look of it, the built-in Django `runserver` development server doesn’t support serving over SSL anyway. Sigh.

PR #48, opened 2022-11-03
Import constituency fuel poverty data

struan:
This also improves the area page data gathering to load all the datasets available for a page rather than hard coding the fetching.


zarino:
FYI @struan I’ve rebased this branch on top of the newly rebased [`33-import-cons-age-data`](https://github.com/mysociety/local-intelligence-hub/tree/33-import-cons-age-data), so that it inherits the new council area page layout too.

hub/templates/hub/area.html
```diff
@@ -151,6 +151,18 @@ <h5 class="card-title mb-3">Age range</h5>
                     </div>
                 </div>
             </div>
+            <div class="col d-flex flex-column">
```
zarino, line 154: I note the source dataset only covers England, meaning the `constituency_fuel_poverty` context variable is empty for non-English constituencies, resulting in something like this:

Screenshot 2022-11-09 at 11 57 34

Maybe we should wrap this entire `<div class="col…">` in an `{% if constituency_fuel_poverty %}` check?

hub/templates/hub/area.html
```diff
@@ -151,6 +151,18 @@ <h5 class="card-title mb-3">Age range</h5>
...
+                <div class="card flex-grow-1 fake-data">
+                    <div class="card-body">
+                        <h5 class="card-title mb-3">Fuel poverty</h5>
```
zarino, line 157: Maybe this title should be "Households in fuel poverty"? Otherwise a "Fuel poverty" figure of, say, "21%" is a bit ambiguous.

alexander-griffen (approved):

zarino (approved):
🚀

struan:
Merged in 35c361db425dd168eb3d99f2ceceecda39bd51ed

PR #49, opened 2022-11-03
Document how to run the tests

zarino:
Realised, today, I had no idea how to run the tests locally. So had a poke around, and tried to document what seems to work for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,3 +42,16 @@ You can create the first Django user by entering a bash shell inside the `web` c
     script/createsuperuser
 
 The superuser will be created with the details specified in the `DJANGO_SUPERUSER_*` environment variables. [Read more about how Docker handles environment variables](https://docs.docker.com/compose/envvars-precedence/).
+
+### Running the tests
+
+First start the Docker environment:
+
+    docker-compose up
+
+Then enter a bash shell inside the `web` container, and run the tests from there:
+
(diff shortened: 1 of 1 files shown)


alexander-griffen (commented):
When I ran this, I had the following error returned:
```
 File "/app/./manage.py", line 11, in main
    from django.core.management import execute_from_command_line
ModuleNotFoundError: No module named 'django'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/app/./manage.py", line 22, in <module>
    main()
  File "/app/./manage.py", line 13, in main
    raise ImportError(
ImportError: Couldn't import Django. Are you sure it's installed and available on your PYTHONPATH environment variable? Did you forget to activate a virtual environment?
```

However, I think this is because the script saved my .env file such that it looked like this:
```
MAPIT_URL="https://mapit.mysociety.org"
MAPIT_API_KEY="secret"DEVENV=docker
```

As when I put the DEVENV to a newline, it worked.

zarino:
@alexander-griffen yeah, I guess that happened because your `.env` file didn’t end with a newline.

alexander-griffen:
> @alexander-griffen yeah, I guess that happened because your `.env` file didn’t end with a newline.

Oh, that makes sense!

alexander-griffen (approved):

PR #51, opened 2022-11-04
Vscode config setup

ajparsons:
This PR fixes some issues with developing *inside* the container (e.g. the vscode approach where it gives you a bash console inside the container, and so all management commands need to work in there).

The key change that may (but shouldn't) affect other aspects is the changes to the Dockerfile. Here I've turned off using virtualenvs inside the docker container (as it's the only thing in there), and this means that running `poetry add [x]` works fine while inside the container. Also involved using the new setup script for poetry and version 1.2.

`script/bootstrap` will now detect it is working inside the docker and change its default to (confusingly) 'native' (because it then is working 'natively' inside the docker when it calls other commands). 

It also adds a github action to mirror back to git.mysociety.org so commits can be made from codespaces straight to github and be available for review. 

End result works fine in codespaces, populates all data successfully.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,20 @@
+
+{
+	"name": "local-intelligence-hub",
+
+	"dockerComposeFile": ["../docker-compose.yml"],
+	"service": "web",
+	"workspaceFolder": "/app",
+	"overrideCommand": true,
+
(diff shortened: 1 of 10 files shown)


struan (approved):
One small tweak to the docker-compose file to make a) tests work and b) non vscode things run but otherwise, looks good.

PR #54, opened 2022-11-07
Adds progress bars to data loading

ajparsons:
Some of these take a while - it's nice to have some feedback on what's happening.

Adds tqdm dependency.

image


struan (approved):

struan:
Might be nice to add either a disable option, or set `disable=None` so it doesn't display if we ever need to run this from a cron.

ajparsons:
@struan : quiet flag (and tests for that), added. Good for me to merge?

PR #58, opened 2022-11-08
Add featured data and non hardcoded area page data

struan:
This allows datasets to be marked as featured for display on the area page.

It also grabs all the data for an area and returns it grouped by category (place, movement etc) and then uses that to display the data. There's a bunch of stuff to work out how to sensibly display the data from properties.

Other small dataset improvements:
* better sourcing data and display of this
* adds percent type so they can be displayed better


zarino (commented):
This all seems to work! (Once I’d re-run the import commands.)

One thing that maybe got lost in translation: The idea we wanted to test out with the "featuring"… errr… _feature_, was whether those "featured" datasets could be the ones that appear by default on an area page, under the MP/Public opinion/Place/Movement category headings. Rather than "featured" datasets being in _their own_ section.

So maybe another way of thinking about it is that, if someone were to visit the site and an admin hadn’t marked _any_ datasets as "featured", then the area pages would be _empty_. (Eventually, they’d be able to _add their own_ datasets to those pages, by "favouriting" them, but that’s a whole other thing.)

Does that make sense @struan ?

If/when you do make that change, we could probably drop all the "fake-data" datasets stuff from the area page too. Might avoid confusion!

zarino (approved):
@struan this looks good to me. I’ve added a commit that changes what’s displayed when there’s no data at all, for a particular category, in the database. Hopefully the commit message explains my logic!

(As a result, the area page looks pretty empty right now—especially if you set `DEBUG=False`—but that problem will solve itself as we start adding more data.)

Feel free to squash any/all of my commits in this branch, if you want.

hub/templates/hub/area.html
```diff
@@ -94,174 +94,61 @@ <h5 class="card-title mb-3">MP profiles</h5>
...
-                </a>
-            </div>
         </div>
 
+      {% if categories.opinion or counts.opinion %}
         <h2 class="mt-5 mb-3 text-primary">Public opinion</h2>
-
         <div class="row row-cols-1 row-cols-md-2 row-cols-lg-3 g-4">
+          {% for category in categories.opinion %}
+            {% include 'hub/area/_place_data.html' with category=category %}
```
zarino, line 103: Oh, one thing – there’s nothing "place" specific in this file. Do you think this should just be `_data.html` or `_datapoint.html` or something, to avoid future confusion?

alexander-griffen (approved):

hub/templates/hub/area.html
struan: I've updated this in a later branch to be `_opinion_data.html` as I'm assuming that they are all going to have their own templates

PR #62, opened 2022-11-16
Import support for renewables polling data

struan:
Fetch and display support for renewables polling data.

At the moment you need to fetch the polling data as a csv from google docs and put that csv in the `data` directory in a file called `renewables_polling.csv` before you run the `import_renewables_polling_data` command.


hub/templates/hub/area.html
```diff
@@ -1,4 +1,3 @@
-{% extends 'hub/base.html' %}
```
zarino, line 1: I think this line has been removed accidentally? It obviously breaks the page layout.

hub/templates/hub/area/_opinion_data.html
```diff
@@ -0,0 +1,14 @@
+{% for item in category.data %}
+    <div class="col d-flex flex-column">
+        <div class="card flex-grow-1">
+            <div class="card-body">
+                <h5 class="card-title mb-3">{{ item.label }}</h5>
+                <p class="card-text mb-0 display-6 lh-1 text-primary">{{ item.value }}%</p>
+                <p class="card-text mt-2 text-muted">{{ item.average|floatformat }}% national average</p>
+            </div>
+            <div class="card-footer bg-white">
+                <p class="card-text fs-8"><a href="#" class="text-decoration-none text-muted">{{ category.source }}</a></p>
```
zarino, line 10: `_place_data.html` says `Data from {{ category.source }}`, why doesn’t this?

hub/templates/hub/area/_opinion_data.html
```diff
@@ -0,0 +1,14 @@
+{% for item in category.data %}
```
zarino, line 1: If we really must have a loop, I’d be tempted to move it out into `area.html`, so that this file works just the same as `_place_data.html`, ie: given a single data item (with a label, value, average, source, etc), it prints it out as a card.

zarino (changes requested):
I was surprised to find these 13 questions listed as a single "dataset" in the Django admin, and then was extra surprised to find, once I set that single dataset as "featured", 13 boxes appeared at once on the area page. Bonk!

I think, instead, we’ll want to treat these as separate "datasets" (or maybe "data points"?) in the admin and frontend, as they’re all about totally different topics, and TCC and/or users might want to "feature" and/or "favourite" just one of the 13 questions, without the other 12 showing up.

(It’s an almost an implementation detail that the questions happen to have come from the same survey – and that should hopefully be acknowledged in the source attribution at the bottom of each card. I can’t see any reason we’d want to "bundle" all these questions up together, aside from that.)

Is that a problem?

zarino (approved):
Looks good to me.

Ironically, in trying to find the correct source for the data, I’ve actually found the _original_ (public) spreadsheet – https://www.survation.com/polling-in-every-constituency-in-britain-shows-strong-support-for-building-wind-farms-to-drive-down-consumer-bills/renewableuk-mrp-constituency-topline/ – don’t know whether you want to update the import script to use that, instead of a local CSV?

PR #64, opened 2022-11-17
Script for running all data import commands in one go

zarino:
Running half a dozen `./manage.py` commands gets really tiresome really quickly. I threw together a quick script to run them all automatically, and figured I may as well share it.

Also includes documentation and some tidying in README.md.

@ajparsons I’ve tried to guess what the process would be in the Visual Studio Code devcontainers, but haven’t tried it out myself, so would appreciate a sanity-check from you that what I’ve written in the README still makes sense. Please comment on anything that needs re-wording.

I’ve left `script/dev-populate` in place, but didn’t want to mention it in the README because it duplicates some (but not all) of the importing. I guess ideally we should either have `dev-populate` run `import-all-data` itself, or remove any import-related stuff from `dev-populate`… or scrap `dev-populate` entirely. @ajparsons any preference?

### DIFF
--- a/script/import-all-data
+++ b/script/import-all-data
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+# abort on any errors
+set -e
+
+# check that we are in the expected directory
+cd `dirname $0`/..
+
+COMMANDS=$(find hub/management/commands/ -name 'import_*' -printf '%f\n' | sed 's#[.]py##')
+
+for COMMAND in $COMMANDS
+do
+    ./manage.py "$COMMAND"
+done
--- a/README.md
+++ b/README.md
@@ -3,9 +3,6 @@
 A Beta verion of a tool to help The Climate Coalition with better access
 to data needed to enable local and national action on climate.
 
-# Contract Countdown
-
-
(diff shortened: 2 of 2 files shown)


ajparsons:
So basically I just want a script that does all the setup steps, so would be happy if `dev-populate` calls that new import script, *or* a different script called the superuser one and then it could be deleted.

script/import-all-data
```diff
@@ -0,0 +1,14 @@
...
+set -e
+
+# check that we are in the expected directory
+cd `dirname $0`/..
+
+COMMANDS=$(find hub/management/commands/ -name 'import_*' -printf '%f\n' | sed 's#[.]py##')
+
+for COMMAND in $COMMANDS
+do
+    ./manage.py "$COMMAND"
```
struan, line 13: I would suggest that this needs to do something to guarantee that `import_areas` is run first as the other commands will not work unless that has run. I suspect explicitly running that first and then do the find thing but with an if so skip if `COMMAND` is `import_areas` is the best way.

struan (changes requested):
One comment on script ordering.

ajparsons:
Completely different approach: should we stop doing lots of management commands in favour of one management command that can recognised defined DataImport classes?

Could then define an import all in python rather than a looping bash script, import classes could have a priority level that solves that areas need to be imported first, etc

zarino:
Pull requests welcome! 😉

struan:
I mean, yes, yes that is sort of the long term plan.

ajparsons:
I just want to backseat drive, don't mind me.

PR #68, opened 2022-11-21
User dataset favouriting

### LINKED ISSUE #31: ‘Favouriting’ datasets to your account
Logged-in users should be able to ‘favourite’ datasets on an area page, that will then appear in a prominent location on all subsquent area page views, until the dataset is ‘un-favourited’.

We may also want to display ‘favourite’ datasets more prominently in the "add data" modal on the map page – but I’ll leave the implementation of that to #29 or one of its spin-off tickets.

The idea we’re testing here is that people will generally only care about a handful of measures/metrics, and they’ll care about that across all the constituencies they look at. So, by ‘favouriting’ the datasets that are most relevant to their job, or their current campaign, we can make sure to display those datasets across the site, and avoid showing _other_ datasets they don‘t care about.

struan:
Screenshot 2022-11-24 at 17 07 34

Fixes #31


struan:
@zarino this is now ready for some design time to make it look nice.

zarino:
@struan I’ve added a fixup commit that sorts out the styling and replaces the markup and javascript with something much simpler.

struan:
Lovely, thanks!

zarino (approved):

alexander-griffen (approved):

PR #70, opened 2022-11-23
import count of wi groups per constituency

### LINKED ISSUE #67: Import WI group location data
The WI website includes a widget for finding your local WI group: 

Looking at the source code, it seems to load its results from a Squiz API:

```https://wi-search.squiz.cloud/s/search.json?collection=nfwi-federations&profile=_default&query=!null&sort=prox&sort=prox&start_rank=1&origin=54.093409,-2.89479&maxdist=<distance>&num_ranks=<results>```

Replacing `<distance>` and `<results>` with large numbers (like, 9999) returns what looks like a list of all WI groups in the UK, 5307 of them.

The data in `response.resultPacket.results` _can_ include a name, postcode, URL, and even meeting day/time for each group … but it looks like only the name is guaranteed, and in limited exploration, I’ve seen at least some groups without a postcode.

struan:
This also extracts out the common code from the GBGW import script rather than have it all duplicated.

Fixes #67

### DIFF
--- a/hub/management/commands/base_importers.py
+++ b/hub/management/commands/base_importers.py
@@ -1,0 +1,73 @@
+from time import sleep
+
+from django.core.management.base import BaseCommand
+from django.db.models import Avg, IntegerField
+from django.db.models.functions import Cast
+
+import pandas as pd
+
+from hub.models import Area, AreaData
+from utils.mapit import (
+    BadRequestException,
+    ForbiddenException,
+    InternalServerErrorException,
+    MapIt,
+    NotFoundException,
+    RateLimitException,
+)
+
+
+class BaseLatLongImportCommand(BaseCommand):
+    def _process_lat_long(self, lat=None, lon=None, row_name=None):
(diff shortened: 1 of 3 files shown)


alexander-griffen (commented):
I haven't been able to run the script all the way through, because it will eventually error out with `requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))`, the first time this is because my internet connection disappeared, but it has happened subsequently - maybe this should be caught?

Also, it may not be worth the time required to implement this at this stage, but it might be nice to have a way of running an update (either for if a new WI group has been posted to the API, or if the script can't be ran all the way through), so that it could be updated without deleting all of the existing data points first.

Otherwise, looks good!

struan:
It probably needs an intermediate step where it takes the API output and sticks in in to a CSV and then goes through the CSV and adding the constituency in which means you could keep running it till you have all the constituencies and then import that.

I might have a go at that next week and replace all this with that.

PR #77, opened 2022-12-01
Add pre_save receiver to cast data into correct data type

gbp:
Add `float` and `int` fields to `CommonData`.

This is needed so we can do proper field lookups against numerical data with the `gte`/`lt` lookups [1].

This automatically migrates the data into the correct column.

[1] https://docs.djangoproject.com/en/4.1/ref/models/querysets/#gt


struan (approved):
Nice.

PR #88, opened 2022-12-07
MP appg import

### LINKED ISSUE #37: Import MP APPG membership data
https://www.parliament.uk/mps-lords-and-offices/standards-and-financial-interests/parliamentary-commissioner-for-standards/registers-of-interests/register-of-all-party-party-parliamentary-groups/registers-published-in-2022/

Note there’s a new list of members every few months.

Would be great to get _all_ APPGs, not just climate/environmental ones. We assume this won’t make the ticket much harder, as all the APPGs have the same page format.

alexander-griffen:
APPG data import, fixes #37 

Relies on a list of climate APPGs which can be accessed by saving [this Google doc](https://docs.google.com/document/d/1AF5eyeHJSGqOALHMSrHDZjSHBvRN--Wm451_iYia7kw/edit?usp=sharing) as a text file. 
(This document was created by manually grepping the APPG names and descriptions for climate-related words. It may be a bit overzealous.)

Similar to the Select Committee memberships, it has a slightly janky method of displaying the APPGs in a view.

The APPGs are displayed in long lists, which doesn't look good with the rest of the layout.

### DIFF
--- a/hub/management/commands/import_mps_appg_data.py
+++ b/hub/management/commands/import_mps_appg_data.py
@@ -1,0 +1,111 @@
+from django.core.management.base import BaseCommand
+
+from mysoc_dataset import get_dataset_df, get_dataset_url
+from tqdm import tqdm
+
+from hub.models import DataSet, DataType, Person, PersonData
+
+party_lookup = {
+    "Conservative": "Conservative Party",
+    "Labour": "Labour Party",
+    "Liberal Democrat": "Liberal Democrats",
+    "Labour (Co-op)": "Labour Co-operative",
+    "Independent": "independent politician",
+    "Alliance": "Alliance Party of Northern Ireland",
+    "Green Party": "Green Party of England and Wales",
+    "Speaker": "Speaker of the House of Commons",
+    "Social Democratic & Labour Party": "Social Democratic and Labour Party",
+}
+
+
+class Command(BaseCommand):
+    help = "Import data on what APPGs an MP is part of"
+
+    source_url = "https://www.parliament.uk/mps-lords-and-offices/standards-and-financial-interests/parliamentary-commissioner-for-standards/registers-of-interests/register-of-all-party-party-parliamentary-groups/"
+    data_url = get_dataset_url(
+        repo_name="mp-appg-membership-data",
+        package_name="mp_appg_membership_data",
+        version_name="latest",
+        file_name="appg_officers.csv",
+    )
+
+    def handle(self, quiet=False, *args, **options):
+        self._quiet = quiet
+        self.import_results()
+
+    def add_mps(self, df):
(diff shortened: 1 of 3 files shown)


hub/management/commands/import_mps_appg_data.py
```diff
@@ -0,0 +1,127 @@
...
+        df = df[df.appg_name.isin(climate_appgs)]
+        df = self.add_mps(df)
+        return df
+
+    def create_data_type(self):
+        appg_membership_ds, created = DataSet.objects.update_or_create(
+            name="mp_appg_memberships",
+            defaults={
+                "data_type": "text",
+                "description": "Membership in APPGs as published on the parliament website",
```
struan, line 72: This should have a `label` and a `source_label` so it shows up in the filters.

hub/management/commands/import_mps_appg_data.py
```diff
@@ -0,0 +1,127 @@
...
+                "source": "https://parliament.uk/",
+                "table": "person__persondata",
+            },
+        )
+
+        appg_membership, created = DataType.objects.update_or_create(
+            data_set=appg_membership_ds,
+            name="appg_membership",
```
struan, line 80: This should be the same name ("mp_appg_memberships") as the dataset otherwise the filtering doesn't work.

hub/templates/hub/area.html
```diff
@@ -159,5 +165,6 @@ <h2 class="mt-5 mb-3 text-primary">Movement</h2>
 
     </div>
 </div>
+<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.1.0/dist/js/bootstrap.min.js" integrity="sha384-cn7l7gDp0eyniUwwAZgrzD06kc/tftFf19TOAs2zVinnD/C7E91j9yyk5//jjpt/" crossorigin="anonymous"></script>
```
struan, line 168: What is this for?

hub/views.py
```diff
@@ -258,6 +258,7 @@ def get_context_data(self, **kwargs):
             ).select_related("data_type")
             for item in data.all():
                 context["mp"][item.data_type.name] = item.value()
+            context["mp"]["appg_memberships"] = [item.value() for item in data.filter(data_type__name="appg_membership")]
```
struan: this will need to change to match the updated data_type__name from the import script.

struan (changes requested):
some small changes around dataset properties but otherwise looks good.

hub/templates/hub/area.html
alexander-griffen: Oops, this is a hold-over from previously messing around with displaying the APPGs. It's been removed, now

struan (approved):

PR #105, opened 2022-12-13
sanity check filter opacity calcs to stop divide by zero errors

struan:


zarino (approved):
Seems fine. Longer term we’ll want to replace this with a colour scale function, rather than opacity.

PR #106, opened 2022-12-14
enable range datasets to be filtered by child data type

### LINKED ISSUE #91: Work out how to handle complex datasets (eg: Constituency age distribution) in Explore page filters
Filtering constituencies by "Constituency age distribution … is less than … 50" doesn’t make sense:

Screenshot 2022-12-08 at 17 02 23

I think there are two ways we could handle this. Either:

1. List all of the sub-datasets (each age group, in this case) as their own separate buttons in the modal picker. So you’d end up with something like:
   Screenshot 2022-12-08 at 17 09 28
2. Keep the dataset as a single button in the modal picker, but add the sub-datasets as _another_ select box once the dataset has been added, before the comparison, like:
   Screenshot 2022-12-08 at 17 08 40

struan:
if a dataset is marked as a range return those as types from the datasets endpoint and then make use of that in the UI. when creating filters also search datatypes if a matching dataset hasn't been found.

Fixes #91


struan:
The JS bit of this seems a bit sub-optimal but I couldn't think of a neater way to do it.

zarino (changes requested):
I’ve added two fixup commits – one which simplifies the javascript, and one that adjusts the CSS to handle the extra `<select>` element.

I can see that the data _type_ (or whatever we call it) gets combined with the comparator and saved to the URL, but _restoring_ from that URL doesn’t work. I think maybe we need some code in `restoreState` to make it aware of data _types_ before the `__`, not just data _sets_? TBH I can’t quite get my head around the code flow from `restoreState` to `loadDatasets` and beyond 😕

Issue #82, opened 2022-12-05
Import Tearfund data on churches that have declared a climate emergency


zarino:
Data’s in [the spreadsheet here](https://drive.google.com/drive/folders/1u5Pdo-mqzAsWnJ46C1jRNXOWqBDaZZER), but will need a bit of cleaning up (eg: moving addresses into the address column).

MVP is being able to see the number of churches, in this constituency, which have declared a climate emergency.

Being able to also see the _names_ / addresses of the churches would be great.

PR #117, opened 2023-01-10
Import churches with climate emergency data

alexander-griffen:
Related to #82 (but doesn't currently show addresses).

Grabs postcodes from first two columns, and converts them to GSS codes in the `generate_tearfund_churches_csv.py`, and then an import script.


struan (approved):
looks good!

Issue #99, opened 2022-12-12
Import Save The Children shop locations


zarino:
Data in [the spreadsheet here](https://docs.google.com/spreadsheets/d/1yY7tLnzrhl-PoybXcdNEfr-cAFOJg2Dx/edit).

Like with the [Tearfund churches](https://github.com/mysociety/local-intelligence-hub/issues/82), I think the MVP is displaying the number of shops in the given constituency.

Being able to see the names / addresses of the shops would be a nice extra step.

PR #118, opened 2023-01-10
Clean + import save the children shops count data

alexander-griffen:
MVP for #99 

Searches for consituencies in the "Constituency" column, but if it fails to find them (either because it's blank, or the constituency doesn't exist, or there's an error), it will use the postcode and MapIt.


struan (approved):
looks good.

zarino:
@struan @alexander-griffen am I right in thinking this hasn’t been deployed (or _has_ been deployed but the import hasn’t been run?)

PR #121, opened 2023-01-12
Select Committee Membership

### LINKED ISSUE #36: Import MP Select Committee membership data
https://committees.parliament.uk/committees/?SearchTerm=&House=Commons&Active=Current

https://committees-api.parliament.uk/index.html

alexander-griffen:
Fixes #36 
Script uses the Parliament Committees API to pull down the MP memberships.
The API appears to allow multiple MPs to be searched in one request, but I couldn't get this to function properly, so it makes a call for each MP individually at the moment - this doesn't seem ideal, but hasn't caused any issues when I've ran it other than taking a bit of a while.

The way this is shown in the view seems a little janky. Is there a better way of implementing it as-is, or should I be configuring the data models differently? Is it worth adding a 'list'(like) datatype?

### DIFF
--- a/hub/management/commands/import_mps_select_committee_membership.py
+++ b/hub/management/commands/import_mps_select_committee_membership.py
@@ -1,0 +1,84 @@
+from django.core.management.base import BaseCommand
+
+import pandas as pd
+import requests
+from tqdm import tqdm
+
+from hub.models import DataSet, DataType, Person, PersonData
+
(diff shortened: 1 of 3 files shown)


struan (changes requested):
One small change.

I think it's ok for now to do the data in the template like this, likewise one call per MP is I think ok.

struan (approved):
This looks good. I wonder if we need a new comparator for these that does a `__contains` lookup so people don't need to know the exact select committee name. Not something for this but I will ticket it.

PR #137, opened 2023-01-18
MP Job titles

### LINKED ISSUE #84: Import MP “job titles” data (ie: cabinet, shadow cabinet positions)
Chris from Green Alliance has shared their working spreadsheet of these, but @struan and I were wondering whether we’ll get better / more timely data from the Parliament API? Or even Wikidata?

alexander-griffen:
Fixes #84 

Just drags them in verbatim, rather than trying to process the textual data.

Currently, it fixes errors in Constituency names manually (as there weren't many), but possibly this could be fixed in the csv itself, as the changes are hard-coded anyway.

### DIFF
--- a/hub/management/commands/import_mp_job_titles.py
+++ b/hub/management/commands/import_mp_job_titles.py
@@ -1,0 +1,86 @@
+from django.core.management.base import BaseCommand
+
+import pandas as pd
+from tqdm import tqdm
+
+from hub.models import Area, DataSet, DataType, Person, PersonData
+
+CONSTITUENCY_CORRECTIONS_DICT = {
+    "Beverly and Holderness": "Beverley and Holderness",
+    "Brighton Pavilion": "Brighton, Pavilion",
+    "Enfield Southgate": "Enfield, Southgate",
+    "Lewisham Deptford": "Lewisham, Deptford",
+    "Ealing Southall": "Ealing, Southall",
+    "Brighton Kemptown": "Brighton, Kemptown",
+    "Richmond": "Richmond (Yorks)",
+    "Na h-Eileanan an Lar": "Na h-Eileanan an Iar",
+}
+
+
+class Command(BaseCommand):
+    help = "Import MP Job titles"
+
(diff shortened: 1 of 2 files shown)


hub/management/commands/import_mp_job_titles.py
```diff
@@ -0,0 +1,84 @@
...
+    help = "Import MP Job titles"
+
+    def handle(self, quiet=False, *args, **options):
+        self._quiet = quiet
+        self.import_results()
+
+    def get_df(self):
+        df = pd.read_csv("data/mp_job_titles.csv", usecols=["Constituency", "Title"])
+        df = df.query(
+            "not(Title.str.contains('Member of Parliament for') | Title.str.contains('MP for'))"
```
struan, line 28: might also be worth stripping "Member of $foo Select Committee" here too, given we're getting that data elsewhere. Worth keeping the Chair bit, unless we can update the select committee importer to pull in that data.

struan (approved):
One small improvement but otherwise this looks good.

zarino (approved):
This looks good to me @alexander-griffen!

I’ve added a fixup commit that tones down the styling, and changes the display name from "Job title(s)" to "Positions", which feels a bit more accurate.

Here’s how it looks with short and long content (although, in the old page layout, of course):

Screenshot 2023-01-25 at 13 38 55

Screenshot 2023-01-25 at 13 39 03

hub/management/commands/import_mp_job_titles.py
zarino: This could be a challenge, if Select Committees are ever listed _alongside_ other useful positions, in the same string. (Does that ever happen?)

zarino: > (Does that ever happen?)

Ah, answered my own question – no it doesn’t (currently).

So yeah, could strip anything that matches `r'^Member.*Select Committee$'`. Feels like keeping the handful of "Chair of […] Select Committee" references is useful, because we don’t currently differentiate between chairs and members of select committees elsewhere on the page.

PR #141, opened 2023-01-19
WIP 98 area page rejig

### LINKED ISSUE #98: Consider combining multiple Public Opinion datasets into fewer “cards”
Currently, the “big percentage” handling of all the different public opinion datasets takes up a lot of space. TCC’s instinct is that combining them into something more akin to the Age bar chart would be easier to scan.

They’ve suggested three thematic groups that might make sense to the typical user:

Card 1:
- Perceived seriousness of Climate Crisis (Onwards polling)
- Support for Net Zero (Onwards polling)
- (Future TCC questions re: Cost of Living etc)

Card 2: 
- Voting intention (Renewables UK)
- Considering voting differently than in 2019 (Renewables UK)
- Less favorable to Conservatives if weakened climate policies (Renewables UK)
- Winning party if Conservatives dropped NV pledge

Card 3:
- Support for renewables all together (Offshore Wind; Onshore Wind; Solar; Tidal; Nuclear - don't need Wave.)
- Wonder whether 'Renewable projects in my area' could go here too? Or is that crossing the streams?

zarino:
Closes #98.

*TODO:*

- Standardise the capitalisation of dataset names (including "Govt"->"government")
- Improve spacing / headers between MP membership list items (@zarino)
- Rename "Support for climate" card header to "Support for net zero"
- APPGs / Select Committees should have a "includes"/"does not include" comparator on the Explore page, with all of the possible options included in drop-down `<select>` element
- Comparators for NZSG and CEN memberships


hub/management/commands/import_foe_groups.py
```diff
@@ -18,6 +18,7 @@ class Command(BaseImportFromDataFrameCommand):
     defaults = {
         "data_type": "integer",
         "category": "movement",
+        "subcategory": "groups",
```
struan, line 21: might be nice to add a data migration to automatically update these.

struan (approved):
This looks good, few tiny comments but nothing to hold this back.

Would be nice to squash it down to slightly fewer commits.

hub/management/commands/import_foe_groups.py
alexander-griffen: I was thinking of just manually changing it in django admin as it isn't a huge amount of changes, is this a bad idea?

hub/management/commands/import_cen_nzsg_members.py
```diff
@@ -20,6 +20,10 @@ def get_df(self):
         return df
 
     def create_data_types(self):
+        options = [
+            {"title": "True", "shader": "#DCDCDC"},
```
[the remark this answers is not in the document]
alexander-griffen: Good question! It's now cyan-700

hub/management/commands/import_foe_groups.py
struan: No, that would also work :)

PR #146, opened 2023-01-25
Fix table error

alexander-griffen:
Fixes error when a dataset with a many-to-one relationship to an Area or Person is viewed in the table view

### DIFF
--- a/hub/tests/test_views.py
+++ b/hub/tests/test_views.py
@@ -48,7 +48,16 @@ class TestPageRenders(TestCase):
 
 
 class TestExploreDatasetsPage(TestCase):
-    fixtures = ["areas.json", "mps.json", "elections.json", "area_data.json"]
+    fixtures = [
+        "areas.json",
+        "mps.json",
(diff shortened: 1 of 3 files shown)


hub/tests/test_views.py
```diff
@@ -59,6 +65,11 @@ def test_explore_datasets_json_page(self):
         response = self.client.get(url)
         self.assertEqual(response.status_code, 200)
 
+    def test_explore_view_with_many_to_one(self):
+        url = f"{reverse('explore_csv')}?mp_appg_membership__exact=MadeUpAPPG"
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
+
```
struan, line 72: might be worth checking that the data returned looks ok too?

struan (approved):
one small test comments but looks good!

hub/tests/test_views.py
alexander-griffen: Good shout! That's been included

PR #153, opened 2023-02-02
Show data_type hints (eg: "%") in Explore page filters

### LINKED ISSUE #92: Explore page: Append "%" symbol to numeric input for percentage datasets
When a dataset’s `data_type` is `percentage`, we should append a "%" symbol to the filter text input, to make it clear that the number is a percentage. So this:

Screenshot 2022-12-08 at 17 11 39

Would instead look like this:

Screenshot 2022-12-08 at 17 14 20

Use [Bootstrap input groups](https://getbootstrap.com/docs/5.2/forms/input-group/) for this.

zarino:
When adding a new filter on the Explore page, the value input will be followed by a "%" if the value must be a percentage, and will trigger a numeric keypad on touchscreen devices if the value must be numeric.

Fixes #92.

Here it is in action, on two percentage datasets, and two integer datasets:

Screenshot 2023-02-02 at 16 13 12

### DIFF
--- a/hub/templates/hub/explore.html
+++ b/hub/templates/hub/explore.html
@@ -31,6 +31,11 @@
                     <select v-if="filter.options" v-model="filter.selectedValue" class="form-select form-select-sm ms-3 flex-grow-1 flex-shrink-1">
                         <option v-for="option in filter.options" :value="option">${ option }</option>
                     </select>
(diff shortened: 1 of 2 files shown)


struan (approved):
👍

PR #154, opened 2023-02-02
Replace hard-coded /status with a proper health check

### LINKED ISSUE #24: Replace /status page with proper health check
At the moment the status page is a basic hard coded template. At some point should replace it with one that at least checks if the database is available.

zarino:
Saw this hanging out at the bottom of the issue list and thought, bah, how hard can it be.

Nice new `/status` endpoint that returns JSON, with a 500 HTTP status code if there was an issue connecting to the database:

```json
{"database": "ok", "areas": 650, "datasets": 35}
```

Turns out it’s slightly messy checking for a database connection in Django. Got it down to 7 or 8 lines. Still, probably simpler than adding a dependency like [django-health-check](https://pypi.org/project/django-health-check/) which would do _way_ more than we need.

Fixes #24.

### DIFF
--- a/hub/views.py
+++ b/hub/views.py
@@ -3,7 +3,9 @@ import json
 from collections import defaultdict
 from operator import itemgetter
 
+from django.db import connection
 from django.db.models import Count, OuterRef, Subquery
+from django.db.utils import OperationalError
 from django.http import HttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404, redirect
 from django.views.generic import DetailView, TemplateView, View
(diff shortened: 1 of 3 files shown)


zarino:
flake8 is unhappy because I used a bare `except` in `get_db_status`. But I’m not sure we really care [which of the Database exceptions](https://docs.djangoproject.com/en/4.1/ref/exceptions/#database-exceptions) it is…?

hub/views.py
```diff
@@ -528,4 +529,28 @@ def get_context_data(self, **kwargs):  # pragma: no cover
...
+        db_status = self.get_db_status()
+
+        http_status_code = 200
+        if db_status == "error":
+            http_status_code = 500
+
+        return JsonResponse(
+            {
+                "database": db_status,
+                "areas": Area.objects.count(),
```
struan, line 552: if the database is down this fails and you get an error page so probably only want to run these if the database is up.

struan (changes requested):
one small bug, but otherwise, yay!

hub/views.py
zarino: Thanks. After your comment I realised I could easily test this out by killing the `db` docker container. So I’ve done that, and simplified the `StatusView` code too.

struan (approved):
👍

PR #155, opened 2023-02-03
Script for Python linting inside or outside Docker container

zarino:
`script/lint` convenience wrapper for our three linting tools. (Because, despite `.vscode/settings.json`, I could never seem to make `flake8` etc available on VSCode’s path to get automated linting to work.)

If a Docker environment is present, it’ll run the linting inside that (respawning itself _inside_ the container if you run it from _outside_). Otherwise, it will assume you want to run natively, inside a Python `poetry` environment.

@ajparsons I’d love your thoughts on this. Not sure if/how it should work with `.vscode/settings.json`.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,10 +3,13 @@
 
     "editor.formatOnSave": false,
     "[python]": {
-        "editor.formatOnSave": true
+        "editor.formatOnSave": true,
+        "editor.codeActionsOnSave": {
+            "source.organizeImports": true
+        }
     },
-
     "python.formatting.provider": "black",
+    "python.formatting.blackPath": "${workspaceFolder}/script/black",
     "files.exclude": {
         "**/.git": true,
         "**/.svn": true,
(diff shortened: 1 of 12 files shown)


ajparsons:
Have had a look and added some extra stuff.

Your version ran fine for me on codespaces, I think I've made the settings.json similarly not dev environment specific, but you'll need to check that for me.

zarino:
@ajparsons `script/lint` still works for me, run from inside or outside the docker container, so I’m happy.

Any reason to have those individual `script/flake8` and `script/black` wrappers, when we could just get VSCode to run `script/dev-command` directly? eg:

```
"python.formatting.blackPath": "${workspaceFolder}/script/dev-command",
"python.formatting.blackArgs": [
  "black ."
],
…
"python.linting.flake8Path": "${workspaceFolder}/script/dev-command",
"python.linting.flake8Args": [
  "flake8 --exclude '.venv' ."
],
```

Another thought, but maybe for a separate commit, is that now we’ve got the generic `script/dev-command`, we could use it inside of `script/test` to allow unit tests and the coverage checker to be run the same way.

ajparsons:
Simple reason: I tried it and it didn't work :( 

My guess is VScode is doing something that means 'blackPath' actually goes to black and is checking for the version or something - so you can't trick it with an extra argument.

ajparsons (approved):
(I'm approving because it works from my pov, but I've also introduced changes so)

zarino:
> Simple reason: I tried it and it didn't work :(

D’oh! That’s a pain. Oh well, a couple of extra files in `script/` isn’t the end of the world.

I’ll see whether I can get `script/test` to use `script/dev-command` too, for simplicity.

struan (approved):
works for me.

PR #158, opened 2023-02-08
Groups UI change

alexander-griffen:
This PR changes how groups are both imported and displayed, such that they will show the groups in a tag cloud with links to the group's URL where possible (instead of a count of those groups).

This consists of these changes:
- Splitting the WI import into a generation + import script
  - Also, changing which information is imported to grab a link and name
- Changing the FoE import script to import names of groups instead of counts
- Changing how this information is displayed in the UI 

To deploy this, the following additional steps will need to be ran:
- Removing the following datasets:
  - Number of Friends of the Earth groups
  - Number of Women's Institute groups
- Rerun the FoE groups import
- Rerun the Women's Institute import
  - (To do this, either the csv will need to be generated with the new command, or the csv can be copied across and just the import command ran, to save time)

### DIFF
--- a/hub/management/commands/generate_wi_groups_csv.py
+++ b/hub/management/commands/generate_wi_groups_csv.py
@@ -1,0 +1,128 @@
+import re
+from time import sleep
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+import pandas as pd
+import requests
+from tqdm import tqdm
+
+from hub.models import Area
+from utils.mapit import (
+    BadRequestException,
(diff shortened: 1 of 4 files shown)


struan (approved):
👍

zarino (approved):
I haven’t run this locally, but the UI should be fine, since it’s just reusing the same component we use further up the page for MP groups/committees.

We’ll just need to keep an eye on the length of the group names – crazy long names might not wrap nicely. But that feels like a bridge we can cross when we encounter it.

Issue #168, opened 2023-02-14
Replace filter select boxes with multi-selects, as a poor man’s “OR” operator


zarino:
Filters on the Explore page are currently treated as “AND” operators – _all_ of the filters must be true for a constituency to appear in the result set, eg, in pseudo-SQL:

```
WHERE `MP party` == "Labour" AND `MP majority` < 5000
```

Notably, that means there’s currently no way in the UI to issue a query like:

```
WHERE `MP party` == "Labour" OR `MP party` == "Labour & Co-operative"
```

Or:

```
WHERE `Select committees` CONTAINS "energy" OR `Select committees` CONTAINS "climate"
```

…Leading people to request some way of adding filters that _expand_ the list of results (“OR”) rather than _contracting_ it (“AND”).

My fear is that a generic “OR” operation between filters will quickly raise questions over precedence and grouping. For example, do we want to allow people to select constituencies `WHERE (x AND y) OR (z)` as well as `WHERE (x) AND (y OR z)` etc? Where do you draw the line … `WHERE (a OR b) AND ((x OR y) AND z)`?

I note that customer-facing interfaces tend to try and avoid these complex nested conditions / combinations. The Smart Playlist configurator in iTunes comes to mind, letting you add multiple filters, but requiring that they either _all_ must be true (our current behaviour, the “AND” route) or that _any_ one or more can be true (essentially all “OR”):

<img width="794" alt="Screenshot 2023-02-14 at 17 55 47" src="https://user-images.githubusercontent.com/739624/218819262-6689eb3f-4b92-4a9f-8aff-8231c10aaaee.png">

Talking this through with Flossie earlier, I started noticing that many (but not all) of the things she’d want to use an “OR” for were represented as `<select>` boxes in the current filtering interface – things like MP Party, or MP select committees. And that made me wonder whether we could just turn those `<select>`s into `<select multiple>`s, to allow implicit “OR” functionality.

By allowing people to select multiple options _within a single dataset_ they could essentially construct queries like:

```
WHERE `MP party` IN ("Labour", "Labour & Co-operative") AND `majority` < 5000`.
```

And because the “OR” is only ever inside a single dataset at a time, there’s no need to worry about grouping or precedence.

What this _wouldn’t_ allow is a query with a condition that spans multiple datasets, like:

```
WHERE `FoE supporters` > 500 OR `WWF supporters` > 500
```

[...]

PR #171, opened 2023-02-15
Allow multiple select filters for e.g political parties

struan:
Part of #168


zarino:
I have pushed a fixup commit that includes those styling improvements.

PR #174, opened 2023-02-16
Improved naming and comparators for MP job titles dataset

zarino:
Using string comparators (rather than default equality comparison) means you can search for all constituencies with an MP whose position contains "minister" or "shadow" or "whip" etc.

I’ve updated the name to include "positions", to match how it’s labelled on the area page.


alexander-griffen (approved):
👍

Issue #143, opened 2023-01-24
Re-arrange public opinion datasets on individual area page


zarino:
We’ve had a couple of suggestions here:

- Move "support for local renewable projects" out of its own big stat card, and into the "renewable energy" table card
- This would make the columns unbalanced, so _also_ move "consider climate change a high priority" out of its own big stat card, and into the "support for net zero" table card
- #142 – even though this potentially conflicts with the above point, now that a totally _separate_ piece of information is in the table, alongside the other options

PR #181, opened 2023-02-23
169 battleground seats

### LINKED ISSUE #169: Party vote split at most recent election in constituency (to help identify “battleground” seats)
We already have MP majority data, which lets us tag constituencies with a majority of less than 5000 as "marginals".

Screenshot 2023-02-14 at 18 09 24

But campaigners will often want to be more specific, and focus down to just marginals where the MP and the runner up were from a given two parties (eg: "Labour/Tory marginals" or "SNP/Labour marginals").

There are a couple of ways we could do this:

1. A **"battleground seat" dataset** for each constituency, which is hand-compiled and a bit of a black box (as it happens, Flossie has already compiled this) but would allow you to, in pseudo-SQL:

   ```
   WHERE `Battleground seat` == "Labour/Conservative marginal"
   ```

   The thing I don’t like about this is that it‘s a dataset _based_ on raw numbers, some of which we _do_ actually hold in the database, but it isn’t actually calculated _from_ those numbers. Feels yucky. And also relies on someone manually updating the list whenever by-elections are held.

2. A **"top two parties" dataset** for each constituency – which breaks out the "majority" aspect of the battleground, and leaves just the party pairing, allowing you to select, for example:

   ```
   WHERE `MP majority` < 5000 AND `Top two parties` == "Labour/Conservative"
   ```

3. A **"second most popular party" dataset** for each constituency – meaning you’d be able to do, eg:

   ```
   WHERE `MP majority` < 5000 AND `MP party` == "Labour" AND `Second most popular party` == "Conservative"
   ```

   …but, [unless we _also_ implemented #168](https://github.com/mysociety/local-intelligence-hub/issues/168), you wouldn’t be able to do this:

   ```
   WHERE `MP majority` < 5000 AND `MP party` IN ("Labour", "Conservative") AND `Second most popular party` IN ("Labour", "Conservative")
   ```

@alexander-griffen @struan any thoughts on the merits / challenges of these approaches?

Related: It feels like importing (and displaying on the area page) the _full_ party vote split for the constituency, at its last election, would also be a useful thing.

Note: Ideally we’d display data from the most recent election in the constituency, which might be a by-election, rather than the 2019 general election.

### Data sources

2019 general election party vote split data is available from UK Parliament here: https://researchbriefings.files.parliament.uk/documents/CBP-8749/HoC-GE2019-results-by-constituency-csv.csv

And UK Parliament also has a spreadsheet of by-elections _since_ the 2019 general election here: https://researchbriefings.files.parliament.uk/documents/CBP-9225/Data-file.xlsx

alexander-griffen:
Fixes #169 

Imports both the party that came second in the last election (filterable), and the results of the last election (not filterable).

### DIFF
--- a/hub/management/commands/import_last_election_data.py
+++ b/hub/management/commands/import_last_election_data.py
@@ -1,0 +1,258 @@
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+import numpy as np
+import pandas as pd
+from tqdm import tqdm
+
+from hub.models import Area, AreaData, DataSet, DataType
+
+
+class Command(BaseCommand):
+    help = "Import data from the last election"
+
+    # https://commonslibrary.parliament.uk/research-briefings/cbp-8749/
+    # https://researchbriefings.files.parliament.uk/documents/CBP-8749/HoC-GE2019-results-by-constituency-csv.csv
+    general_election_source_file = (
+        settings.BASE_DIR / "data" / "2019_general_election.csv"
+    )
+    # https://commonslibrary.parliament.uk/research-briefings/cbp-9225/
+    # https://researchbriefings.files.parliament.uk/documents/CBP-9225/Data-file.xlsx
+    by_election_source_file = settings.BASE_DIR / "data" / "byelections.xlsx"
+
+    party_translate_down_dict = {
+        "Conservative": "con",
+        "Labour": "lab",
+        "Green": "green",
+        "Liberal Democrats": "ld",
+        "Liberal Democrat": "ld",
+        "Social Democratic": "sdlp",
+        "Conservative Party": "con",
+        "Labour Party": "lab",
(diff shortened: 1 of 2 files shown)


alexander-griffen:
@zarino I'm struggling with the display of this - both getting the table to display in order of votes cast, and also getting the date to display.

zarino:
@alexander-griffen I’ve removed the "second place" card from the area page (no need, given there’s a vote split table instead) and moved the table into a different column where there’s likely to be a bit more space.

It’d be nice to have a little pie chart here, or something, but I figure we leave that to be addressed as part of the future improvements in https://github.com/mysociety/local-intelligence-hub/issues/143.

Screenshot 2023-02-24 at 17 09 46

zarino (approved):
This looks good to me. Just needs linting.

PR #188, opened 2023-02-28
CSV Data export

### LINKED ISSUE #179: Management command for generating a spreadsheet of (almost) all the data in the Hub
We’d like a management command we could run, to export as much constituency-level data as possible out of the Hub, for testing / experimenting with.

The output should be a single spreadsheet/CSV, with one row per constituency, and one column per “dataset”. We could probably limit it to any datasets that are is_filterable: true, if that makes it easier (avoiding complicated JSON data types).

Eventually we’ll be able to [add custom columns to the CSV export on the Explore page](https://github.com/mysociety/local-intelligence-hub/issues/93), but even then, I expect this command may remain useful as a convenient way to get hold of an archive of _all_ the data, without adding columns one by one.

alexander-griffen:
Fixes #179 

This command generates a csv file of all of the filterable datasets in the app.

What's interesting is that it reveals some inconsistencies in strategies for data imports. It is possible to alter the script so that it hides those consistencies a little better, but I haven't been sure if it's better to do that, or use it as a tool to identify the inconsistencies and figure out how to improve them in the imports themselves. (This is particularly noticable with how I have approached booleans in different places throughout the app).

(Currently, I've left the commits as separate WIP commits to more clearly show changes that have been made, but I'll squash them together before a merge)


zarino (approved):
YES YES YES 🎉

Issue #203, opened 2023-03-16
Inconsistent requirement for 'updated button'


ajparsons:
In this example, if you change '70' to '50' and hit 'table view', it will update the table, but it will not update the url parameter.

![image](https://user-images.githubusercontent.com/8157058/225562016-afa8cd7e-ec14-41da-8370-dad1216dea1e.png)

This means if you *now* press the 'csv download', you get the '70' dataset, not the 50. You need to press 'update results' before the csv download is correct. 

If there has been a change to the config, hitting one out the output formats should be the 'same' as hitting 'update results'. Not clear currently when data is 'stale'.

PR #221, opened 2023-03-29
add additional column support for explore data

### LINKED ISSUE #93: Explore page: Allow user to add extra columns to the Table and CSV output
When the user selects Table or CSV output, we include a column for the constituency name, and a column for each filter (but not shader!) dataset that has been applied.

TCC has suggested that users are likely to want to include _more_ columns – datasets that aren’t being used to filter the list, but will be useful once the table is copied / downloaded, for further analysis or combination with other datasets.

We could use the existing dataset modal UI markup/JS for this. Just need a way for people to invoke it, if Table or CSV output is selected.

I’m thinking maybe the order of the sidebar should be:

1. Add filters
2. Select output format
3. (…output format options)
4. Update results (button)

Where the third section changes based on the selected output format. For Map output (the default) we’d show the shader widget. For Table and CSV output, we’d show an "Additional columns" widget.

struan:
Fixes #93

### DIFF
--- a/hub/fixtures/area_data.json
+++ b/hub/fixtures/area_data.json
@@ -8,6 +8,7 @@
       "last_update": "2022-10-10T13:20:30+00:00",
       "is_range": "True",
       "source": "https://parliament.uk",
+      "table": "areadata",
       "category": "place"
     }
   },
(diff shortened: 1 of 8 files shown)


zarino:
@struan I’ve added some commits here that change the UI and also fix #203.

I notice one thing we’re missing (which I _know_ people will ask for immediately) is the ability to add the MP’s name as a column on the table.

struan:
I'll have a think about the MP name but I can hard code the MP name at least as a temporary measure.

zarino (approved):
This looks good to me, @struan :+1:

PR #224, opened 2023-04-05
119 improve shaders

### LINKED ISSUE #119: Explore page: Improve shader scaling and presentation
Shaders currently:

- Use a scale based on the min and max values from the entire dataset (min and max values generated at dataset import time, not map query time).
- Use only the opacity of polygons to represent the percentage.

Instead, shaders should:

- Use a scale based on the minimum and maximum values from _the constituencies matching the current filters_, so the visible constituency on the map with the lowest value in the chosen shader dataset will always have a "shade" of 0%, and the visible constituency on the map with the highest value will always have a "shade" of 100% – even if those constituencies would actually be somewhere in the middle of the pack if you looked at all 650 constituencies as a whole. In other words, the scale shouldn’t be based on all 650 constituencies, it should only be based on the constituencies matching the current filter(s).
- Use a colour scale, [like Viridis or Magma](https://cran.r-project.org/web/packages/viridis/vignettes/intro-to-viridis.html) and keep fill and stroke opacities at 100%.

Not for the MVP, but a nice-to-have would be:

- Display a key/legend somewhere, to explain the scale, showing the entire scale as a gradient, with the minimum and maximum values at either end.

struan:
Fixes #119

### DIFF
--- a/hub/models.py
+++ b/hub/models.py
@@ -240,13 +240,30 @@ class DataSet(TypeMixin, models.Model):
 
     def colours_for_areas(self, areas):
         values, mininimum, maximum = self.shader_value(areas)
-        props = {
-            "properties": {
-                "maximum": maximum,
-                "minimum": mininimum,
-                "shades": self.shades,
+        legend = {}
+        for option in self.options:
+            if option.get("shader", None) is not None:
+                legend[option["title"]] = option["shader"]
+
+        if len(legend) > 0:
+            props = {"properties": {"legend": legend}}
+        else:
+            d_max = maximum
+            d_min = mininimum
+            if self.is_float:
+                d_max = round(maximum, 1)
+                d_min = round(mininimum, 1)
+                if self.is_percentage:
+                    d_max = f"{d_max}%"
+                    d_min = f"{d_min}%"
+
+            props = {
+                "properties": {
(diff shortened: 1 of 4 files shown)


hub/models.py
```diff
@@ -199,53 +199,120 @@ class Meta:
...
+        for option in self.options:
+            if option.get("shader", None) is not None:
+                legend[option["title"]] = option["shader"]
+
+        if len(legend) > 0:
+            props = {"properties": {"legend": legend}}
+        else:
+            props = {
+                "properties": {
+                    "maximum": maximum,
```
zarino, line 251: I think these values are what gets passed through to `explore.html` in `${ key.minimum }` and `${ key.maximum }` … if so, is there any way of pre-formatting these (rounding to a sensible number of decimal places, and adding a `%` sign if `data_type == percent`?

I considered doing the formatting inside the vue.js code, but then realised that generating a human-readable miniumum and maximum for a dataset feels like something the model should handle itself?

If `properties.minimum` and `properties.maximum` need to be maintained as they are, then maybe an _additional_ `properties.minimum_formatted` and `properties.maximum_formatted` (or similar) could be added instead?

struan: why, yes, yes there is.

PR #226, opened 2023-04-06
Alter MP relevant votes import

### LINKED ISSUE #185: Fix MP activity/support datasets, to always show on MP’s page, and show if MP wasn’t in post at the time

alexander-griffen:
Fixes #185 

Now, when MPs weren't in office for the vote/EDM, it shows as such. This also means that the vote is always displayed on the MP pade

### DIFF
--- a/hub/management/commands/import_mps_relevant_votes.py
+++ b/hub/management/commands/import_mps_relevant_votes.py
@@ -27,6 +27,7 @@ class Command(BaseCommand):
         edms = self.get_all_edms()
 
         self.data_types = self.create_data_types(votes, edms)
+        self.delete_data()
         self.import_results(votes, edms)
 
     def get_votes(self, division_id):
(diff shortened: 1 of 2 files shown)


hub/management/commands/import_mps_relevant_votes.py
```diff
@@ -104,10 +138,14 @@ def create_data_types(self, votes, edms):
         vote_options = [
             {"title": "Aye", "shader": "#89c489"},
             {"title": "No", "shader": "#e18674"},
-            {"title": "No vote", "shader": "#fede86"},
+            {"title": "Did not vote", "shader": "#fede86"},
+            {"title": "Did not vote (teller)", "shader": "#cff4fc"},
+            {"title": "Not in office", "shader": "#073a41"},
```
struan, line 143: We should probably get @zarino to suggest some non red/green colours here for accessibility reasons.

struan (approved):
One teeny point about colours aside, this looks good.

hub/management/commands/import_mps_relevant_votes.py
alexander-griffen: Thank you! :D

PR #229, opened 2023-04-12
Google Analytics page and event tracking

### LINKED ISSUE #198: Analytics and event tracking
- [ ] Set up a Google Analytics property and make sure TCC team have access (should we set it up in _their_ Google Analytics organisation?)
- [ ] Add cookie-less GA4 tracking script to the site (use CAPE as an example)
- [ ] Generate GA4 events when:
   - homepage area/MP search form submitted
   - homepage geolocation link clicked
   - explore page "add data to filter by" button clicked
   - explore page filter added (bonus points if we can record the dataset name, to track popularity of datasets over time)
   - explore page "add dataset to shade results" button clicked
   - explore page shader added (again, bonus if we can record dataset name)
   - explore page map, or table, or CSV rendered
   - explore page map constituency clicked
- [x] Update [privacy policy](https://github.com/mysociety/local-intelligence-hub/issues/69) to acknowledge cookie-less Google Analytics tracking

zarino:
- Get Google IDs etc from environment vars
- analytics.esm.js module with promise-based tracking
- tracking for various JavaScript events

Fixes #198.


zarino:
@struan when you get a moment, could you just double-check that all the environment variable stuff here will work ok on the staging and live sites? (I tried to copy what we did in CAPE, but LIH’s setup is a bit different.) Guess we could merge and then find out, but feels like it’d be nice to have one single clean commit that adds everything we needed – if only so we can look back at it in future and work out what we need to do for other similar projects!

PR #231, opened 2023-04-17
Record datetime of last pageload on UserProperties model

zarino:
Part of https://github.com/mysociety/climate-project/issues/87.

First commit adds a `last_seen` field to the `UserProperties` model, into which the datetime of each subsequent pageload is saved. This way, we can see when a user was last active on the site.

The `last_seen` data isn’t displayed anywhere, yet, but is at least accessible via the admin UI if we need it.

Speaking of which, the second commit adds more columns, search, and filtering options to the UserProperties list in the admin UI:

Screenshot 2023-04-17 at 14 23 39

@struan I have broken the tests, because they create a user without any `userproperties` (which wouldn’t happen in real usage, because users _should_ only be created via the account request form, and that creates the `user.userproperties`) – should we fix the tests to always create a `userproperties` for the test users? Or should we make `RecordLastSeenMiddleware` silently handle missing `user.userproperties`?


struan:
I've pushed a fix to this that only updates the last seen if we've not seen them today, and also stores this in the session which should cut out a chunk of DB access and still retain enough information to be useful here.

PR #238, opened 2023-04-19
Terms of use page and checkbox during signup

zarino:
Adds a "Terms of use" page, and a checkbox for it in the signup form.


zarino:
Yeah, hard-coding the URL is fine! I’ll get this rebased and merged :+1:

PR #239, opened 2023-04-19
Import National Trust property locations

### LINKED ISSUE #127: National Trust properties data
Could we get it from Wikipedia / Wikidata? https://en.wikipedia.org/wiki/List_of_National_Trust_properties_in_England

Feels like that would be simpler than scraping it from the National Trust website – https://www.nationaltrust.org.uk/visit

struan:
Fixes #127

### DIFF
--- a/hub/management/commands/generate_national_trust_properties_csv.py
+++ b/hub/management/commands/generate_national_trust_properties_csv.py
@@ -1,0 +1,132 @@
+import json
+from time import sleep
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+import pandas as pd
+from tqdm import tqdm
+
+from hub.models import Area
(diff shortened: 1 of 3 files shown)


zarino:
I’m going to merge and deploy this, as it looks good on the Area page, and seems to work fine as a filter on the Explore page. But just noting here, there seems to be a bug when this is used as a shader:

Screenshot 2023-04-21 at 09-38-00 Explore Local Intelligence Hub

Far too many constituencies are being shaded dark blue, suggesting they have, what, 15, 16 properties each? Liverpool Riverside, for example, is showing as dark blue, but when you actually click on the constituency and look at the area page, only one property is shown (The Hardmans’ House). As a local, I know this is correct – there aren’t any other NT properties in this constituency.

Issue #192, opened 2023-03-09
Store and display release dates for all datasets


zarino:
- [x] Check the sources are correct, and consistently capitalised, phrased, etc.
- [ ] Append dataset release date to all dataset source labels.

zarino:
Part of this (consistent naming) was done in https://github.com/mysociety/local-intelligence-hub/pull/240.

Going to move the remaining part (dataset release dates) into the next milestone, as it’ll take a little bit of time to research.

zarino:
The technical aspect of this was done in #288. We are collecting release dates in #307.

zarino:
What’s left to do here is for someone (@alexander-griffen?) to [go through the spreadsheet here](https://docs.google.com/spreadsheets/d/1OdvrAML9KBlAaWERtCd1MkbMC_3I1B0MHTd7GRKFoJk/edit), and copy all the dataset publication dates over into the relevant import scripts.

Complications:

- Dataset release date is currently set as a DateField in the Django model – it should probably just be free text (sometimes we might have a specific date, sometimes it might be a month and year, sometimes it might just be a year, sometimes it might be "fuzzy" like "Autumn 2023", who knows!)
- Some import scripts will need to _set their own_ release date, as they scrape live data, so the release date should be the date the scraper was last run.

PR #240, opened 2023-04-19
192 dataset names and sources

zarino:
Part of #192.

- Standardise wording for dataset names.
- Add data sharing warning to MRP datasets on area page.

@struan, once this is merged, is there a script we can run to update all the dataset names on the live site, or should I go in and edit them by hand via the Django admin?

### DIFF
--- a/hub/management/commands/import_area_countries.py
+++ b/hub/management/commands/import_area_countries.py
@@ -32,7 +32,7 @@ class Command(BaseCommand):
             defaults={
                 "data_type": "text",
                 "description": "The country that the constituency is in",
-                "label": "Country",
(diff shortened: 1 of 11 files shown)


struan (approved):
TBH other that re-importing them all then changing them in the django admin is the only other option and re-importing takes some time so editing in django is as quick. I'd need to update the re-import script to skip some of the data we don't want as well.

zarino:
👍 I’ll merge this and then apply the dataset name changes manually via the Django admin.

zarino:
> 👍 I’ll merge this and then apply the dataset name changes manually via the Django admin.

Done and done.

PR #248, opened 2023-04-25
Replace activate_accounts with a full accounts page

zarino:
- `/activate_accounts` -> `/accounts`
- Now displays full table of users (combining properties from `auth.User` and `UserProperties` models)
- Also displays counts of active and joined users in last 7 and 30 days

I decided not to display the users’ email addresses A) to save space, and B) to slightly reduce the chance of personal data getting accidentally screenshared at some point ;-) The email is still accessible via the name link, which leads to the Django Admin page for that `userproperties` instance.

Screenshot 2023-04-25 at 09-28-30 Accounts Local Intelligence Hub


struan (approved):
I've pushed a couple of small fixes to the status code, plus a redirect from the old URL for the sake of tidyness.

PR #249, opened 2023-04-27
MP changes no longer causes multiple MPs to an area

### LINKED ISSUE #225: MP changes
When a new MP starts, re-running the MP import script will import the new MP without removing the old MP, which leads to 2 MPs for a constituency which will break things.

That's said, if MPs are deleted every time the script is run, all PeopleData will get dropped.

Possibly, a solution should ensure that relevant data imports are re-ran (at least for the new MP) when a new MP is imported

alexander-griffen:
Fixes #225

When the import_mps command is ran, if a new MP had been elected in that time, it would add them to the database without removing the previous MP. This means that when you navigate to the relevant constituency webpage, it would return a MultipleObjectsReturned error.

Now, the script will identify multiple MPs to an area, and then keep only the most recently elected one. It will then re-run all of the MP data scripts, so that the information for the MP is populated.

Currently, the way that the `PersonData` imports are identified is by hard-coding the command names into a constant. Possibly, it might be a better solution to search for filenames with a specific prefix (e.g. `mp_data_datasetname`), but this could cause unnecessary confusion in future development.

Additionally, the script will run the _entire_ commands, not just for a specific MP (particularly because many of them will read from a CSV, or pull data from an API in bulk. It's possible, but might be unnecessary, to refactor some of these so that an argument can be passed to the command that will return the data for only that MP.

### DIFF
--- a/hub/management/commands/import_mps.py
+++ b/hub/management/commands/import_mps.py
@@ -3,7 +3,9 @@ from collections import Counter
 
 from django.conf import settings
 from django.core.files import File
+from django.core.management import call_command
 from django.core.management.base import BaseCommand
+from django.db.models import Count
 
 import magic
 import requests
@@ -28,6 +30,14 @@ party_shades = {
     "independent politician": "#DCDCDC",
 }
 
+MP_IMPORT_COMMANDS = [
+    "import_mps_relevant_votes",
+    "import_mps_select_committee_membership",
+    "import_mps_appg_data",
+    "import_mp_job_titles",
+    "import_mp_engagement",
+]
+
 
 class Command(BaseCommand):
     help = "Import UK Members of Parliament"
@@ -40,6 +50,7 @@ class Command(BaseCommand):
     def handle(self, quiet: bool = False, *args, **options):
         self._quiet = quiet
         self.import_mps()
+        self.check_for_duplicate_mps()
         self.import_mp_images()
 
     def get_mp_data(self):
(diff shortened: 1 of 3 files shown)


hub/management/commands/import_mps.py
```diff
@@ -41,6 +50,7 @@ def handle(self, quiet: bool = False, *args, **options):
         self._quiet = quiet
         self.import_mps()
         self.import_mp_images()
+        self.check_for_duplicate_mps()
```
struan, line 53: I might put this ahead of `import_mp_images` as then it's a) not fetching images for the old MP and b) the fetch bit takes a while so it reduces the time where the "two records found" error potentially appears.

hub/management/commands/import_mps.py
```diff
@@ -200,3 +210,36 @@ def import_mp_images(self):
             image = File(open(file, "rb"))
             mp.photo.save(f"mp_{mp.external_id}.{extension}", image)
             mp.save()
+
+    def check_for_duplicate_mps(self):
+        areas_covered = list(Person.objects.values_list("area", flat=True))
+        duplicate_areas = [
+            area for area in set(areas_covered) if areas_covered.count(area) > 1
+        ]
+        if duplicate_areas != []:
```
struan, line 219: FWIW this can be done in one in the database although it's not really worth changing here given how infrequently this runs but this would also work:

``` python
duplicates = (
    Person.objects.distinct()
    .values("area_id")
    .annotate(area_count=Count("area_id"))
    .filter(area_count__gt=1)
)
if duplicates.count() > 0:
```

plus changing `area=area` to `area=area["area_id"]` in the filter further down.

struan (approved):
One small execution order thing but this looks good.

We should think about if we want to do this in a way that maintains the previous results in future by flagging people as `not_current` or similar but that's outwith the scope of this.

hub/management/commands/import_mps.py
alexander-griffen: This is much cleaner than my implementation, so I've put it in.

PR #250, opened 2023-04-27
Standardise categorical data shading

### LINKED ISSUE #230: Categorical data shading
Standardise the colours for the shaders for categorical data.

See [this comment by Zarino](https://github.com/mysociety/local-intelligence-hub/pull/226#discussion_r1164119878) for colours from the `/style` page that have good contrast.

Paraphrasing a slack message from Zarino for more information:
>We could probably standardise on those 6–8 colours as the go-to set for all the categorical datasets, unless the datasets require something different (like party colours), which means fixing up the existing import scripts

alexander-griffen:
Fixes #230 

- Standardises colours in existing imports
- Adds `is_shadable` option to DataSet models, and updates Shader modal so only shadable datasets can be selected

When merging this, run `./manage.py migrate` to run the migrations which will add the `is_shadable` option to the model, as well as update existing data in the database.


alexander-griffen:
@zarino Some imports aren't feasible to use with the shader (for example, the APPG dataset, because it doesn't have a 1:1 relationship between datapoints and area, and even if it did, has too many categorical options to be a useful form of data visualisation), so would it be useful to add an option to turn the ability to show up on the the shader dropdown off?

zarino:
@alexander-griffen yeah, sounds like we want an `is_shadeable` equivalent to `is_filterable`.

hub/migrations/0037_dataset_is_shadable.py
```diff
@@ -0,0 +1,18 @@
...
+        ("hub", "0036_userproperties_agreed_terms"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="dataset",
+            name="is_shadable",
+            field=models.BooleanField(default=True),
+        ),
+    ]
```
struan, line 18: would be good to add a data migration too, a la b602469, to update the relevant datasets rather than re-running imports

hub/management/commands/import_area_fuel_poverty_data.py
```diff
@@ -33,6 +33,7 @@ class Command(
                 "source_type": "xlxs",
                 "data_url": data_url,
                 "table": "areadata",
+                "is_shadable": False,
```
struan, line 36: I think this is ok to filter by?

struan: although we're not using this dataset anymore in favour of the EFPC one.

hub/migrations/0037_dataset_is_shadable.py
struan: And maybe also to do the colours from the previous commit?

struan (approved):
This looks good.

One suggestion for a data migration to save re-running a bunch of imports but otherwise yay!

hub/management/commands/import_red_blue_wall_constituencies.py
```diff
@@ -26,8 +26,8 @@ class Command(BaseImportFromDataFrameCommand):
         "table": "areadata",
         "is_filterable": True,
         "options": [
-            dict(title="Red Wall", shader="red"),
-            dict(title="Blue Wall", shader="blue"),
+            dict(title="Red Wall", shader="red-500"),
+            dict(title="Blue Wall", shader="blue-500"),
```
struan, line 30: I must have something missing on my side because these both show up as black.

struan (changes requested):
Actually, one display issue for me.

struan:
I've pushed a fix to a bug that was causing some of the shading to break for things like red wall/blue wall

hub/management/commands/import_area_fuel_poverty_data.py
alexander-griffen: You're right! I've removed it, either way

hub/management/commands/import_red_blue_wall_constituencies.py
alexander-griffen: This is in part because I omitted a dictionary lookup that allowed this to work. I have added that back in. Hopefully, this will work better, now

struan (approved):
👍

PR #260, opened 2023-05-04
Cost of living polling data

alexander-griffen:
Fixes #222 

This PR achieves the following:
 - Imports cost of living data.
 - Also extracts the opinion table template into a partial, to avoid unnecessary repetition
 - Corrects a spelling mistake: 'comissioned' -> 'commissioned'

PR #263, opened 2023-05-11
Country exclusions for datasets

### LINKED ISSUE #247: Handle setting data to 0 for datasets that don't apply to all countries
e.g NT property counts don't apply in Scotland so setting those to 0 doesn't really make sense as it will skew the numbers.

Suspect need a `countries` property on the dataset and if it's present then only fill empty countries listed. Or maybe an exclude countries list as many of the datasets are everywhere but Scotland.

alexander-griffen:
Fixes #247.

Adds `exclude_countries` and `fill_blanks` parameters to `DataSet` model, and updates relevant import scripts to use these.

Updated import scripts also have the new data migrated over.


alexander-griffen:
**UPDATE:** Definitely require a sanity check, but I believe that the correct scripts have been set to exclude the correct countries.

Currently, this PR lacks a migration to save having to re-import all of the datasets.

Note that when this is done, #260 and #259 will need to be rebased on top, and have the appropriate country exclusions added to them.

struan (approved):
I've had a quick look over this and it seems fine and certainly does the right thing locally.

PR #267, opened 2023-05-18
skip shading if filters return no areas

### LINKED ISSUE #266: explore.json crashes if select shader when filters return no areas
e.g. request: `?party__in=Conservative+Party&mp_election_majority__lt=15000&constituency_cc_high__gte=55&would-change-party__gte=35&hnh_mrp_29_should__gte=60&ages_30-39__gte=50&shader=would-change-party](https://www.localintelligencehub.com/explore.json?party__in=Conservative+Party&mp_election_majority__lt=15000&constituency_cc_high__gte=55&would-change-party__gte=35&hnh_mrp_29_should__gte=60&ages_30-39__gte=50&shader=would-change-party`

```
TypeError at /explore.json
type NoneType doesn't define __round__ method
...
File "/data/vhost/[www.localintelligencehub.com/local-intelligence-hub/hub/models.py](http://www.localintelligencehub.com/local-intelligence-hub/hub/models.py)", line 274, in colours_for_areas
    d_max = round(maximum, 1)
```

Probably should just skip shading if no areas.

struan:
Display a message in place of the key/scale to make it clear.

This avoids an error when trying to round the None value returned for max/min if there are no areas.

Fixes #266

### DIFF
--- a/hub/models.py
+++ b/hub/models.py
@@ -257,6 +257,9 @@ class DataSet(TypeMixin, models.Model):
         return self.shades[shade]
 
     def colours_for_areas(self, areas):
+        if len(areas) == 0:
+            return {"properties": {"no_areas": True}}
+
         values, mininimum, maximum = self.shader_value(areas)
         legend = {}
         for option in self.options:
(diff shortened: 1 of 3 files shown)


zarino (approved):
@struan I’ve added a fixup commit that prevents `this.key` or `this.legend` from being set if `this.no_areas` is true, which avoids a bug where there was extra space below the no areas message (due to `v-if="key"` still evaluating to true in the template).

Feel free to squash and merge!

PR #277, opened 2023-06-14
Socio-economic data import

### LINKED ISSUE #145: Import socio-economic data
Some users have requested a way to see and filter by the sorts of jobs that people in a constituency have – typically called "socio-economic status".

https://commonslibrary.parliament.uk/find-the-socio-economic-status-of-people-living-in-england-and-wales-by-constituency/ (seems to be england and wales only – any equivalent for the whole UK?)

alexander-griffen:
Fixes #145

Mostly done, however a question remains about the labels of the data.

Ideally, I'd like to have it word-for-word from the data source. However, some of the labels exceed the character limit. To solve this, I altered them slightly, and included the full name in the description.

However despite this, it still displays jankily, as shown in the screenshot. Would it be better to more aggressively alter the names? Would this alter the integrity of the stats?
image

### DIFF
--- a/hub/management/commands/import_socio_economic_status_data.py
+++ b/hub/management/commands/import_socio_economic_status_data.py
@@ -1,0 +1,118 @@
+from django.core.management.base import BaseCommand
+
+import pandas as pd
+from tqdm import tqdm
+
+from hub.models import Area, AreaData, DataSet, DataType
+
+
+class Command(BaseCommand):
+    help = "Import data about socio-economic statuses"
+
+    source_url = "https://commonslibrary.parliament.uk/find-the-socio-economic-status-of-people-living-in-england-and-wales-by-constituency/"
+    data_url = "https://data.parliament.uk/resources/constituencystatistics/PowerBIData/Census%202021/NS-SEC_2021.xlsx"
+
+    message = "Importing constituency socio-economic data"
+    defaults = {
+        "label": "Socio-Economic Status",
+        "description": "Socio-economic status as of the 2021 census",
+        "data_type": "percent",
+        "category": "place",
+        "source_label": "House of Commons Library",
+        "is_range": True,
+        "source": source_url,
+        "source_type": "xlxs",
+        "data_url": data_url,
+        "table": "areadata",
+        "exclude_countries": ["Scotland", "Northern Ireland"],
+        "comparators": DataSet.numerical_comparators(),
+        "default_value": 10,
+        "is_shadable": False,
+    }
+
+    machine_readable_names = {
+        "Higher managerial, administrative and professional occupations": "higher_professional_administrative_occupations",
+        "Lower managerial, administrative and professional occupations": "lower_professional_administrative_occupations",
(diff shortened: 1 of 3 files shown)


hub/management/commands/import_socio_economic_status_data.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        data_types = []
+        for col in tqdm(df.columns, disable=self._quiet):
+            if col != "gss":
+                data_type, created = DataType.objects.update_or_create(
+                    data_set=data_set,
+                    name=f"se_{self.machine_readable_names[col]}",
+                    defaults={
+                        "data_type": "percent",
+                        "label": self.label_names[col],
```
struan, line 84: could maybe do `self.label_names.get(col, col)` here to avoid having to include the ones that aren't being overridden? and then maybe rename it `label_names_override`.

hub/management/commands/import_socio_economic_status_data.py
```diff
@@ -0,0 +1,128 @@
...
+        "source": source_url,
+        "source_type": "xlxs",
+        "data_url": data_url,
+        "table": "areadata",
+        "exclude_countries": ["Scotland", "Northern Ireland"],
+        "comparators": DataSet.numerical_comparators(),
+        "default_value": 10,
+    }
+
+    label_names = {
```
struan, line 32: Might be worth a comment here to explain these are because of character length limits.

Or we could avoid this and just update the table to increase the label length. There's no reason it's set to 200 other than I arbitrarily picked that as probably long enough.

struan: wouldn't solve the display issues though which is a secondary thing.

struan (changes requested):
one tiny config change, and some thoughts about labels.

hub/management/commands/import_socio_economic_status_data.py
alexander-griffen: Label changes now has the display looking like this:
image
Which obviously looks better, but I'm not sure if the label-wrangling is okay from a data integrity point of view?

alexander-griffen: As a table, with no more label mangling:
image

hub/management/commands/import_socio_economic_status_data.py
```diff
@@ -0,0 +1,128 @@
...
+        "source_label": "House of Commons Library",
+        "is_range": True,
+        "source": source_url,
+        "source_type": "xlxs",
+        "data_url": data_url,
+        "table": "areadata",
+        "exclude_countries": ["Scotland", "Northern Ireland"],
+        "comparators": DataSet.numerical_comparators(),
+        "default_value": 10,
+    }
```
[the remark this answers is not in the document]
struan: gah, sorry, I mean `is_shadable` :|

struan (approved):
One tiny tweak because I am a fool.

zarino (approved):

PR #278, opened 2023-06-15
Dataset annotation in preparation for constituency boundary changes

### LINKED ISSUE #254: Annotate datasets with information for future combining
To prepare for automatic translations, for each dataset, store if the data is best expressed as:

1. A raw unit (people, buildings) or a percentage.
2. If it is more correct to see that unit or percentage as being evenly distributed over ‘people in an area’ or ‘the physical area’.

[This ticket is part of the [new constituency data](https://docs.google.com/document/d/144CO0QeVzVbNLFtCafB2rNl54AqHJZtfI3oAL0oFMpk/edit#) plan]

alexander-griffen:
Fixes #254

- [x] Add fields for the type of data, as well as the kind of distribution of unit per data
- [x] Create data migration for existing datasets
    - [x] Create migration for all 'straight forward' datasets in db
    - [x] Figure out how to handle more unusual AreaData datasets in this regard (e.g. `text` or `json` data)
    - [x] Update migration when 'point' data handled
- [x] Handle 'point' data (see ticket)
- [x] Update LIH tests for new fields in admin


struan (approved):
👍

Issue #280, opened 2023-06-20
Only allow logged-in users to see “members only” datasets on area, explore, and data source pages


zarino:
- [x] Enable public access to all pages of the site, add privacy field to dataset model
   - https://github.com/mysociety/local-intelligence-hub/pull/392

Below moved to #399
- [ ] Update all datasets to set privacy field – either public or private
- [ ] Will require us to mark datasets as public or private – [TCC have started a spreadsheet to collect this information](https://docs.google.com/spreadsheets/d/10nCxPl6-HPSel6gLQmhLWOfAWr0rImC6gXX2_lYzPjE/edit#gid=0) from partners / partner agreements [update 4/12/23]

emilyk383:
Will require us to mark datasets as public or private – [TCC have started a spreadsheet to collect this information](https://docs.google.com/spreadsheets/d/10nCxPl6-HPSel6gLQmhLWOfAWr0rImC6gXX2_lYzPjE/edit#gid=0) from partners / partner agreements.

Issue #210, opened 2023-03-16
Reconsider Area page layout, to handle more datasets in future


zarino:
The current Area page layout is already groaning under too many datasets, requiring users to do lots of scrolling around, and making it harder to pick out the dataset you’re interested in.

The current layout is also _heavily_ stage-managed, and with each new dataset we add to the site, we have to rejig the boxes on the Area page to make room.

This is okay (ish) for an initial launch. But we should spend some time thinking about the best way to solve this in the longer term.

alexander-griffen:
In terms of the structure of the templates, the partials are mostly decided by the (human-understandable) context of the data (e.g. is it Place data, or about Opinion, etc?). It might be easier to separate them out based on how the data is displayed (Just a stat, a stat with an average, or table or graph, etc etc). This would also reduce duplicate code across the templates.

If this would make the redesigning of the area page layout easier, I'd be happy to pick it up

PR #285, opened 2023-06-27
Member's only datasets backend

struan:
This is the database and admin part of member's only datasets that enables a dataset to be marked as public.

By default all datasets are private.

Part of #280


alexander-griffen (approved):

zarino (approved):
I‘ve rebased this onto a more recent `main`.

While we’re still awaiting confirmation from TCC, it’s very likely there will be at least _some_ public opinion data and movement data outside the members-only wall. So I’ve tweaked the `area.html` conditionals to reveal more content. Again, it’s annoying that we’re having to fiddle with the markup on this page so much (and that it’s disconnected from the `is_public` flag in the database) but hopefully I can resolve all of that via #210.

I‘ve also fixed some missing nav bar items for logged-out users, and added two prompts to make it clearer that logged-out users are missing out on members-only data.

PR #287, opened 2023-06-28
Display filter values in map tooltips

### LINKED ISSUE #207: Show numeric values for currently applied filters, when hovering over constituencies on the map
We currently show the constituency name in a tooltip when hovering over a constituency on the Explore page map.

Beta testing suggested that, when filters are applied, it would be useful to show the current value of each filtered dataset, in that tooltip, when you hover.

So, for example, you might be filtering by "Support for net zero > 60%" – and hovering over each constituency would show the constituency’s name, along with "Support for net zero: 62%" or whatever that constituency’s value is.

struan:
Fixes #207

### DIFF
--- a/hub/static/js/explore.esm.js
+++ b/hub/static/js/explore.esm.js
@@ -291,6 +291,7 @@ const app = createApp({
               }
             },
             onEachFeature: (feature, layer) => {
+              layer.name = feature.properties.name
               layer.bindTooltip(feature.properties.name)
               layer.on({
                 mouseover: (e) => { e.target.setStyle({ weight: 5 }) },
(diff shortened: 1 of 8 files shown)


hub/static/js/explore.esm.js
```diff
@@ -341,12 +342,21 @@ const app = createApp({
             if ( features[layer.feature.properties.PCON13CD] ) {
               var props = features[layer.feature.properties.PCON13CD];
               // "show" the feature
+              var text = "<strong>" + props["name"] + "</strong><br>"
+              for (p in props) {
+                if (["name", "type", "color", "opacity", "PCON13CD"].includes(p)) {
+                  continue
+                }
+                text += p + ": " + props[p] + "<br>"
```
zarino, line 350: If `props[p]` is a float, we should format it to a sensible number of decimal places…

Screenshot 2023-07-06 at 17 15 38

And we should also show "%" after it, if it’s a percentage.

zarino: More generally, I wonder whether we could override the styling in this tooltip, to make it feel more like the rest of the site – `font-size: var(--bs-body-font-size)` and `font-family: var(--bs-body-font-family)` on `.leaflet-tooltip` might do it?

zarino: Finally, I wonder whether a table layout in here might make it easier to scan down the list of numbers? I’m just thinking, when you have a few datasets active, it’s like a game of hide and seek trying to parse the numbers from each line, whereas a table layout would force the numbers into a neat column down the right-hand side.

zarino (changes requested):
Generally good. I’ve left a couple of comments about tooltip content styling.

zarino (approved):
@struan I’ve added a fixup commit with some styling improvements. I think this is good to go from my perspective.

It’d be nice if the table _also_ appeared in the tooltip when you’re hovering over constituencies that _don’t_ match the given filters (so you could see _why_ they’ve not matched) but I have a feeling that will be a much bigger job that requires a change in how `explore.json` returns features. So maybe something to tackle as a separate change.

struan:
I've added ticket #297 to capture adding the same data to excluded areas.

Issue #205, opened 2023-03-16
Make context clearer when choosing and configuring filters on Explore page


zarino:
Testing showed that people found some filters difficult to use a first, because they were unfamiliar datasets and the UI presents very little context about them. "What’s a sensible number to put in this box?". "I just want to see above average constituencies." etc

I wonder whether a way to fix this would be to present more context next to / as part of the filter controls. We already store min, max, and average values, but we don’t present them.

An additional improvement would be to add “above average” and “below average” options to the default numeric comparators!

zarino:
Potential UI for this:

![Apr-03-2023 16-28-24](https://user-images.githubusercontent.com/739624/229556550-82e8fc88-7698-42e9-83d8-e0f24736fd2c.gif)

```html
<details class="mt-3">
  <summary class="text-muted fs-8 fw-bold lh-1">About this dataset</summary>
  <p class="text-muted fs-8 mb-0 mt-2">mySociety’s Composite UK IMD Index, allowing comparisons across England, Scotland, Wales, and Northern Ireland. 1 = lowest decile, most deprived; 10 = highest decile, least deprived.</p>
</details>
```

zarino:
Closed by #288

PR #288, opened 2023-06-29
provide more context on suitable values for numeric filters

struan:
Set the default to the average and include the max and min in the info box, if available.

Part of #205


struan:
@zarino this could use a bit of design expertise.

zarino (approved):
👍 I’ve gone with the fold-out `<summary>` markup I suggested in #205, and while I was there, I made the dataset description and source visible too.

Now that dataset descriptions are more prominent in the UI, we’ll want to go through and make sure all the import scripts either set the descriptions to something useful (ie: not just a repetition of the dataset title) or leave the description blank. But that’s a task for a separate ticket.

PR #289, opened 2023-06-29
link from area page tags to filtered explore page

### LINKED ISSUE #209: Tags on Area page should link to Explore page filtered by that tag
eg: Clicking "Red Wall" tag on an Area page should load the Explore page map, filtered by "Red or Blue wall == Red wall".

struan:
Everything is hard coded because the criteria is also essentially hard coded.

Fixes #209


alexander-griffen (approved):
🎉

PR #290, opened 2023-07-03
281 update to django 4.2

### LINKED ISSUE #281: Update to Django 4.2
4.2 is the Long Term Support version.

struan:
Fixes #281


alexander-griffen (approved):
This all works nicely - although I'm not sure if the update will require dropping existing data as it did for me locally.

struan:
It only needs the db dropping/re-creating in docker because we had to upgrade from postgres 11 as django 4.2 doesn't support that. On the server we're not touching the db.

PR #292, opened 2023-07-06
display feedback if filters don't return any data

### LINKED ISSUE #268: give better feedback when filters result in no areas being selected
At the moment all that happens is you get a blank map which isn't terribly helpful. There should probably be some sort of feedback to indicate when you filter out all areas.

struan:
Fixes #268

### DIFF
--- a/hub/static/css/_explore.scss
+++ b/hub/static/css/_explore.scss
@@ -1,3 +1,9 @@
+$osm-sea-blue: rgb(168, 212, 243);
+
+[v-cloak] {
+    display: none !important;
+}
+
 .explore-filters {
     overflow: auto;
     border-bottom: 1px solid $border-color;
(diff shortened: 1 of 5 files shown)


struan:
Screenshot 2023-07-06 at 14 58 41

zarino (approved):
@struan I‘ve added a commit that changes quite a few bits of this PR, most notably moving the status display up into a bar across the top of the "results" half of the screen (feels most natural, given that’s where you look to see the output of the modifications you’ve made on the other side):

some-areas

no-areas

I’ve also sorted out a working "loading" spinner, which engages any time `explore.json`, `explore.csv`, or `exploregeometries.json` are being loaded:

loading

Hopefully the commit message makes it all clear! I don’t know whether you want to squash your commits into mine, or leave your work and my work as two separate commits? I think combining the whole lot into one makes the most sense.

struan:
I've pushed a small fix to maintain the Loading... bit in the filter select dialog. I think I will make this two commits though as it makes more sense to me like that.

PR #302, opened 2023-07-21
Django Admin for Area, AreaData, DataType, Person, PersonData models

zarino:
When I attempted to diagnose a bug with missing MP Majority data, I realised there was no easy way to see which data points were actually in the database. This commit fixes that.

- Display more of the models in the Django Admin.
- Display related DataTypes (ie: sub-DataSets, like age bands in the age distribution dataset) on the DataSet admin page.
- Give items in the Area model a human-readable name (the area name) so they’re easier to recognise in the Django Admin.

### DIFF
--- a/hub/admin.py
+++ b/hub/admin.py
@@ -1,6 +1,14 @@
 from django.contrib import admin
 
-from hub.models import DataSet, UserProperties
+from hub.models import (
+    Area,
+    AreaData,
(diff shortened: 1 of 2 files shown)


struan (approved):
Pushed one small fix to the person and area data list displays but otherwise good 👍

PR #309, opened 2023-08-01
DATASET: Air pollution data

### LINKED ISSUE #180: Air quality / pollution data
From Mary:

> Finding usable data on air pollution is proving tricky. Global Action Plan hosts this https://www.cleanairhub.org.uk/forecasts but it is live changing data, pulling info from the [government defra/met hub](https://uk-air.defra.gov.uk/forecasting/locations?q=+M50+3AZ), which works with patterns and weather predictions. What I can't see is the data defra is using to make the predictions?

Beth shared a link to [addresspollution.org](https://addresspollution.org/), which says of its source data:

> Imperial College London’s Environmental Research Group (ERG), who are the UK’s leading authority on air pollution. With their sophisticated modelling techniques and network of state of the art air pollution sensors, combined with the monitor data sourced and mapped by COPI from every UK council, ICL have provided us with a national air pollution model accurate to 20m sq.
>
> addresspollution.org is using the latest available annualised data from 2019. When newer data becomes available, the site will be updated.

It’d be great to find an up-to-date source of air quality / pollution data per constituency.

alexander-griffen:
Fixes #180 

## Implementation notes
Data import of various air pollution data models Defra have produced (that I assume are the basis for some of the data mentioned in the ticket). This data was recorded in 2021.

The geo-matching was performed in GIS, and uploaded to the Drive (explained in more detail in 934f547).

Subscript is currently handled as follows:
- In a dropdown, it's in plain-text shorthand 
image

- When displayed in a table, the subscript is rendered using a custom filter
image

## Next steps
The table (image above) could include more detail on the metric. Below is it as a separate column to show the relevant data (which is stored in the `description` attribute of the `DataType` object), although that column isn't present in the code, because it looks bad. The subscript and superscript is present as unicode here, so shouldn't require fancy handling.
image


zarino (approved):
@alexander-griffen Hope you don’t mind, I’ve added a commit that changes how this data is presented in the front end. Should probably squash my commit into your second commit, before merging.

Issue #256, opened 2023-05-02
Support multiple geographies


ajparsons:
Add support for multiple geographies to LiH, initially to set up an extra map for local authorities, but this is also creating a framework to plug the new constituencies into. 

More details in the doc.

[This ticket is part of the [new constituency data](https://docs.google.com/document/d/144CO0QeVzVbNLFtCafB2rNl54AqHJZtfI3oAL0oFMpk/edit#) plan]

- [x] import new constituencies
- [x] enable swapping area types on explore page
- [x] display new boundaries on explore page
- [x] #334
- [x] #335
- [x] #336
- [x] #337
- [x] #360

alexander-griffen:
## Update
Done so far:
- Adding the `AreaType` model is ready to review [(PR)](https://github.com/mysociety/local-intelligence-hub/pull/310)
Next steps:
- Reviewing [prep work document](https://docs.google.com/document/d/1975tU3XU7LGTFLHeF4vnrylhaz3bDffE9RNplF4dmOQ), and comments. Updating to reflect these changes
- Implementing changes to `DataSet` and `DataType` models to accomodate multiple `AreaType`s

zarino:
Done!! 🎉

PR #310, opened 2023-08-03
Add the AreaType model

alexander-griffen:
This PR adds the AreaType model, creates an object for the 2010 constituency boundaries, and then backfills the current Area objects with the new AreaType object. This is split into several migrations to allow for not-null constraints.

This PR does not totally complete #256, but is the first step towards doing so. Currently, it makes no visible changes to the website.


struan (approved):
yay for multi-step migration process 🎉

struan:
Merged in db4a60a3600cca0b54572aa4c2c21ca2e397c009

PR #311, opened 2023-08-03
Flake8 E721 fixes

alexander-griffen:
Uses `isinstance(x, y)` over `type(x) == y`

I'm not sure when this cropped up, and indeed doesn't seem to get covered by the version of flake8 inside the Docker image.


struan (approved):
👍 🍦

PR #312, opened 2023-08-14
Members only datasets frontend

### LINKED ISSUE #280: Only allow logged-in users to see “members only” datasets on area, explore, and data source pages

struan:
This does three things:
 - hides non public datasets from unauthenticated users
 - updates the area page layout a bit to tidy up the resulting empty boxes etc for unauthenticated users
 - turns off the mandatory login for all pages

Fixes #280

### DIFF
--- a/hub/fixtures/area_data.json
+++ b/hub/fixtures/area_data.json
@@ -10,7 +10,8 @@
       "is_range": "True",
       "source": "https://parliament.uk",
       "table": "areadata",
-      "category": "place"
+      "category": "place",
+      "is_public": "True"
     }
   },
   {
--- a/hub/fixtures/elections.json
+++ b/hub/fixtures/elections.json
@@ -7,7 +7,8 @@
       "data_type": "integer",
       "last_update": "2022-10-10T13:20:30+00:00",
       "source": "https://wikipedia.org",
-      "table": "person__persondata"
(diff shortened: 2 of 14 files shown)


zarino:
@struan I’ve pushed a fixup commit that adds a yellow section to the bottom of the area page, advertising the extra data for logged in members:

Screenshot 2024-01-17 at 09 49 20

Migrations hell prevented me actually testing this out, so I did the work in `main` and then cherry-picked into this branch, guessing that `{% if not user.is_authenticated %}` was the right thing to check in the template – please replace if not :-)

(FYI longer term, I think we’ll split this yellow section into two columns, and use one of the columns to host the email signup form required by #377.)

zarino:
Closing this PR as we have #392 now instead.

PR #313, opened 2023-08-24
DATASET: Flood risk data

### LINKED ISSUE #195: Flood risk data
There are a number of flood risk datasets on https://www.data.gov.uk/search?q=flood+risk, but they all seem to be Shapefiles 🤔

alexander-griffen:
Fixes #195 

Flood risk data csv produced using qgis, file can be found in shared google drive.

Also, a change to the _place_data template that means that datasets with no actual data in them do not render empty cards


hub/templates/hub/area/_place_data.html
```diff
@@ -1,131 +1,152 @@
...
+                            {% endfor %}
+                        </tbody>
+                    </table>
+                {% elif category.name == 'Flood risk from rivers or sea' %}
+                        <table class="table mb-0 js-chart" data-chart-type="bar" data-chart-direction="y"">
+                                <thead>
+                                    <tr>
+                                        <th scope="col">Risk</th>
+                                        <th scope="col" data-color="#068670">This area</th>
+                                        <th scope="col" data-color="#ced4da">UK average</th>
```
struan, line 116: This should be England not UK?

hub/management/commands/import_flood_risk_data.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+class Command(BaseCommand):
+    help = "Import flood risk data"
+
+    data_file = settings.BASE_DIR / "data" / "risk_of_flooding.csv"
+    message = "Importing constituency flood risk data"
+    defaults = {
+        "label": "Flood risk from rivers or sea",
+        "description": "Data relating to risk of flooding from rivers or sea, recorded in 2018",
```
struan, line 17: probably worth adding something to these to make it clear it's for england only, maybe just add "in England" after sea, or "for England" after Data.

struan (approved):
Some small suggestions to make the area it applies to clear.

Would be good to have documentation somewhere of how the data was derived and what it means. My assumption here is that for each constituency there's a per $some-sub-area flood risk rating that we are aggregating into these percentages but that's only a guess from how we're turning the data into numbers.

hub/templates/hub/area/_place_data.html
zarino: This is an issue for a number of datasets that are England or England+Wales only. I figure it’s something we chalk up to fix later, rather than holding up this dataset import.

PR #316, opened 2023-09-11
Button for staff to download CSV of user accounts

zarino:
"Download as CSV" button on the `/accounts` page:

Screenshot 2023-09-11 at 15 31 01

Producing a CSV of all registered accounts, including email addresses.


hub/views/accounts.py
```diff
@@ -98,3 +100,42 @@ def get_context_data(self):
...
+            "date_last_seen",
+            "email_confirmed",
+            "account_confirmed",
+            "is_active",
+            "is_staff",
+            "is_superuser",
+        ]
+        writer = csv.DictWriter(response, field_names)
+        writer.writeheader()
+        for userprops in self.object_list:
```
struan, line 126: if you have a `context_object_name = "users"` at the top of the class then this can use `context["users"]` instead of `self.object_list` which would, I think, be preferable.

struan (approved):
One tiny style thing but otherwise yay!

PR #331, opened 2023-10-26
Multiple Geographies

alexander-griffen:
So far, this PR:

- Downloads the new geographies
- Alters the Area pages to better handle them

What is required is:

- Allowing the map and table on the explore page to handle the two sets of geographies - with a switch?
- Making the area page View less inefficient (probably by moving the lookups to be downloaded to the Area model?)

For #256


hub/management/commands/import_new_constituencies.py
```diff
@@ -0,0 +1,49 @@
...
+import pandas as pd
+from tqdm import tqdm
+
+from hub.models import Area, AreaType
+
+
+class Command(BaseCommand):
+    help = "Import basic area information for new constituencies"
+
+    data_file = settings.BASE_DIR / "data" / "new_constituencies.csv"
```
zarino, line 13: Would be nice to add a comment here, to remind us where this anonymous CSV came from. I’m pretty sure it was https://geoportal.statistics.gov.uk/datasets/ons::westminster-parliamentary-constituencies-future-boundaries-ew-bfc/about

hub/management/commands/import_new_constituencies.py
```diff
@@ -0,0 +1,49 @@
...
+            "-q", "--quiet", action="store_true", help="Silence progress bars."
+        )
+
+    def handle(self, quiet: bool = False, *args, **options):
+        if not quiet:
+            print("Importing Areas")
+        df = pd.read_csv(self.data_file)
+        area_type, created = AreaType.objects.get_or_create(
+            name="2023 Parliamentary Constituency",
+            code="WMC23",
```
zarino, line 26: It looks like we’ve made a decision, elsewhere (eg: https://pages.mysociety.org/2025-constituencies/) to call these "2025" constituencies. So should we use `2025 Parliamentary Constituency` and `WMC25` here? /cc @ajparsons

hub/management/commands/import_new_constituencies.py
zarino, line 13: Would be nice to add a comment here, to remind us where this anonymous CSV came from. I’m pretty sure it was https://geoportal.statistics.gov.uk/datasets/ons::westminster-parliamentary-constituencies-future-boundaries-ew-bfc/about

hub/management/commands/import_new_constituencies.py
```diff
@@ -0,0 +1,49 @@
...
+        parser.add_argument(
+            "-q", "--quiet", action="store_true", help="Silence progress bars."
+        )
+
+    def handle(self, quiet: bool = False, *args, **options):
+        if not quiet:
+            print("Importing Areas")
+        df = pd.read_csv(self.data_file)
+        area_type, created = AreaType.objects.get_or_create(
+            name="2023 Parliamentary Constituency",
```
zarino, line 25: It looks like we’ve made a decision, elsewhere (eg: https://pages.mysociety.org/2025-constituencies/) to call these "2025" constituencies. So should we use `2025 Parliamentary Constituency` and `WMC25` here? /cc @ajparsons

hub/models.py
```diff
@@ -388,6 +388,17 @@ def shader_value(self, area):
         return None, None, None
 
 
+class AreaType(models.Model):
```
zarino, line 391: PS @alexander-griffen I learned a few weeks ago that you can pass a string to `models.ForeignKey()` when you want to refer to a model that’s defined _further down_ in the file. So, in this case, a line like…

```python
area_type = models.ForeignKey("AreaType", on_delete=models.CASCADE, null=True)
```

…would have saved you having to shuffle this Class further up the file.

hub/templates/hub/area.html
```diff
@@ -39,10 +39,40 @@ <h1 class="m-0">{{ area.name }}</h1>
                 </div>
             </div>
             <div class="col-lg-9 offset-xl-1">
-
+            {% if area_type == "WMC" %}
+            <div class="card flex-grow-1 dataset-card">
```
zarino, line 43: It’d be good to have one of these cards on the `WMC23` pages too, but going the _other_ way (ie: saying "This is a 2025 (2023?) constituency … it replaced the following 2010 constituency/ies: …"

hub/templates/hub/area.html
```diff
@@ -39,10 +39,40 @@ <h1 class="m-0">{{ area.name }}</h1>
...
+                                        {% endfor %}
+                                    </tbody>
+                                </table>
+                                </div>
+                            </div>
+            {% endif %}
                 <section id="mp" class="area-section">
                     <h2 class="mb-3 text-primary">MP</h2>
-                    {% include 'hub/area/_mp_data.html' with mp_data=mp %}
+                    {% if area_type == "WMC23" %}
```
zarino, line 64: We’ll probably be replacing the "MP" section with a "Candidates" section, in the run-up to the election. But for now, I’d be tempted to just hide the MP section (and its corresponding link in the subnav) entirely, on WMC23 pages.

zarino (commented):
@alexander-griffen I‘ve cherry-picked the five useful commits out of your original branch and onto `main`, so the PR diff looks much cleaner now! (Your original branch has been backed up to [`256-multiple-geographies-backup`](https://github.com/mysociety/local-intelligence-hub/compare/256-multiple-geographies-backup) in case we ever need it again.)

The "This is a 2010 constituency" card at the top of the Area page looks fine.

I’ll sort out a constituency generation switcher on the Explore page next week.

In the meantime, I‘ve left a few questions / comments inline.

hub/management/commands/import_new_constituencies.py
```diff
@@ -0,0 +1,49 @@
...
+            description="Westminster Parliamentary Constituency boundaries, as created in 2023",
+        )
+
+        for index, area in tqdm(df.iterrows(), disable=quiet):
+            a, created = Area.objects.get_or_create(
+                gss=area["PCON25CD"],
+                name=area["PCON25NM"],
+                area_type=area_type,
+            )
+            geom = area[
```
struan, line 37: this should be the polygon for the area. Off the top of my head I don't know where we get this from. I suspect we're better off not importing this at the moment and having the new constituencies only visible as area pages till we can find the data.

hub/management/commands/import_new_constituencies.py
```diff
@@ -0,0 +1,49 @@
...
+                    "BNG_E",
+                    "BNG_N",
+                    "LAT",
+                    "LONG",
+                    "Shape__Area",
+                    "Shape__Length",
+                    "GlobalID",
+                ]
+            ].to_dict()
+            a.geometry = geom
```
struan, line 48: This irksomely produces bad JSON so we'd need to do `a.geometry = json.dumps(geom)`

struan (changes requested):
Mostly looks fine. one comment about the geometry stuff which is hampered by lack of data.

struan:
A further issue is that GSS codes turn out not be unique across the two different boundary sets so we need to update import scripts etc to distinguish areas based on GSS/Name and area type.

I am going to add some methods to the Area model for looking up areas so we can consolidate this in once place.

zarino (commented):
@struan @alexander-griffen I’ve added a commit that fixes a few of the comments I left last week. The other open comments are all beyond my ability.

Generally, this seems ok. It’d be nice to get those outstanding comments addressed though.

hub/templates/hub/area.html
struan: I'm going to fix this as part of #337

hub/management/commands/import_new_constituencies.py
zarino: FWIW I discussed this with @ajparsons and we think we should refer to these new constituencies as either "2025 constituencies" or "Future constituencies" (if we want to hedge our bets) when we’re limited for space, and "Constituencies at the next election" if we have more space.

I see @struan has already modified `import_new_constituencies.py` to use the latter :+1:

PR #344, opened 2023-11-09
Respect area type in filters

struan:
This handles both the back end task only filtering on available filters for the area type, and also the front end of indicating when a selected filter type is not available when switching area types.

This also updates the filter list to indicate if a filter is available for the current area type.

Screenshot 2023-11-09 at 16 22 33


zarino (changes requested):
I’ve added a fixup commit that simplifies the styling for the "unavailable" filters (no need to show the form controls, if they’ll have no effect). I guess this same styling / markup will need to be replicated for the shaders in the `shaderContainer`, and the columns in the `columnContainer` (which currently has no `areas_available` logic at all).

I‘ve also added a commit that generates the list of area type `<option>`s from data on the JavaScript side, rather than being hard-coded in the HTML markup. Felt like that’s more along the lines of the way the rest of the logic on this page works – but feel free to revert it, if it’s not suitable.

One more thing that needs fixing, I think, is that datasets should only appear in the "Add dataset" or "Add shader" modals if they apply to the current area type. No sense in displaying datasets that people can’t/shouldn’t choose. We’ll need to do this filtering _before_ the `favouriteDatasets`, `featuredDatasets`, and `otherDatasets` arrays are passed to `add-data-modal.html`, so that any "empty" arrays are properly skipped.

zarino (approved):
Bit tricky to test this out locally as I don’t have any WMC23-only datasets, but from what I can see, this looks good to go :+1:

struan:
merged into #350

PR #345, opened 2023-11-14
Import overlapping constituencies

### LINKED ISSUE #337: import and store overlapping old/new constituencies
At the moment we're fetching this every time so it's slow. Need to import this and store it.

Alternatively, we should remove this functionality until we have time to do it more efficiently.

struan:
Fixes #337


zarino (approved):
Looks good to me @struan. Can confirm the migration and import script worked for me in the Docker container, and things look right in the UI.

I’ve added a fixup commit that changes the markup a little in `area.html` – with fresh eyes, I was annoyed by how the table layout tended to break each constituency name onto two lines, and the `text-muted` paragraph after each constituency still felt too prominent. So I’ve gone with a grid layout now, which is optimised for 2–3 replacement constituencies but will handle more or fewer than that too. I’ve also tweaked the wording in the special case where a constituency is a straight 1-for-1 swap, now saying that the constituency is "replaced" rather than being "divided".

Screenshot 2023-11-21 at 18 13 24

Screenshot 2023-11-21 at 18 13 31

Screenshot 2023-11-21 at 18 13 38

Screenshot 2023-11-21 at 18 13 44

PR #348, opened 2023-11-16
DATASET: Save the children supporters

### LINKED ISSUE #197: Save The Children supporter data
Number of Save The Children supporters in each constituency.

See `2022_Campaigning_Actions_By_Constituency 1.csv` file in the shared drive.

alexander-griffen:
Fixes #197 

Notes:
- Due to the fact that the "release_date" changes haven't been merged yet, that line in the defaults is commented out for now
- I manually altered the csv to correct Cosntituency names - but the original is still available in the Drive, suffixed with (OLD) in the title

### DIFF
--- a/hub/management/commands/import_save_the_children_supporter_data.py
+++ b/hub/management/commands/import_save_the_children_supporter_data.py
@@ -1,0 +1,53 @@
+from django.conf import settings
+
+import pandas as pd
+
+from hub.models import AreaData, DataSet
+
+from .base_importers import BaseImportFromDataFrameCommand
+
+
+class Command(BaseImportFromDataFrameCommand):
+    help = (
+        "Import data about the number of Save the Children supporters per constituency"
+    )
+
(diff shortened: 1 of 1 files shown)


hub/management/commands/import_save_the_children_supporter_data.py
```diff
@@ -0,0 +1,51 @@
...
+        "category": "movement",
+        "subcategory": "supporters_and_activists",
+        "source_label": "Data from Save the Children.",
+        # "release_date": "2022",
+        "source": "https://www.savethechildren.org.uk/",
+        "source_type": "csv",
+        "data_url": "",
+        "table": "areadata",
+        "default_value": 10,
+        "comparators": DataSet.numerical_comparators(),
```
struan, line 35: Ideally we should have the unit_type and unit_distribution in here given it won't get backfilled by a migration.

struan (approved):
One small comment but otherwise good.

PR #350, opened 2023-11-21
multiple geometry staging

struan:
This is purely a staging branch for all the multiple geometry stuff to go into once reviewed so we don't have a load of long running interdependent PRs


zarino:
Noting for the new year – I’m getting a 500 error on the staging site, when I try to load the area page for a new constituency:

```
Exception Type: TypeError at /area/WMC23/Bridlington and The Wolds
Exception Value: 'AreaData' object is not iterable
Raised during: hub.views.area.AreaView
```

PR #355, opened 2023-11-28
do some validation on filters before updating the map

### LINKED ISSUE #328: Null value in an int filter results in 500 error
If a user doesn't fill in a value in a in filter then you end up with an error:

```
Exception Type: ValueError at /explore.json
Exception Value: Field 'int' expected a number but got 'null'.
```

which is in `hub/filters.py`, line 24

To replicate select any numeric filter, remove the value and then click  "Update map"

struan:
Highlight empty filters and also prevent the map updating if there are empty filters.
Also ignore empty filters at the back end to avoid 500 errors

Fixes #328

### DIFF
--- a/hub/templates/hub/explore.html
+++ b/hub/templates/hub/explore.html
@@ -47,11 +47,11 @@
                             <option v-for="option in filter.options" :value="option">${ option }</option>
                         </select>
                         <div v-else-if="filter.data_type == 'percent'" class="input-group flex-grow-1 flex-shrink-1">
-                            <input type="text" inputmode="numeric" pattern="[0-9]*" v-model="filter.selectedValue" class="form-control form-control-sm text-end">
+                            <input type="text" inputmode="numeric" pattern="[0-9]*" v-model="filter.selectedValue" class="form-control form-control-sm text-end" :class="{ 'is-invalid': filter.selectedValue === '' }">
                             <span class="input-group-text lh-1">%</span>
                         </div>
-                        <input v-else-if="filter.data_type == 'integer'" type="text" inputmode="numeric" pattern="[0-9]*" v-model="filter.selectedValue" class="form-control form-control-sm flex-grow-1 flex-shrink-1">
-                        <input v-else type="text" v-model="filter.selectedValue" class="form-control form-control-sm flex-grow-1 flex-shrink-1">
+                        <input v-else-if="filter.data_type == 'integer'" type="text" inputmode="numeric" pattern="[0-9]*" v-model="filter.selectedValue" class="form-control form-control-sm flex-grow-1 flex-shrink-1" :class="{ 'is-invalid': filter.selectedValue === '' }">
+                        <input v-else type="text" v-model="filter.selectedValue" class="form-control form-control-sm flex-grow-1 flex-shrink-1" :class="{ 'is-invalid': filter.selectedValue === '' }">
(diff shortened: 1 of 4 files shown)


hub/templates/hub/explore.html
```diff
@@ -47,11 +47,11 @@ <h3 class="h6">${ filter.title }</h3>
...
+                            <input type="text" inputmode="numeric" pattern="[0-9]*" v-model="filter.selectedValue" class="form-control form-control-sm text-end" :class="{ 'is-invalid': filter.selectedValue === '' }">
                             <span class="input-group-text lh-1">%</span>
                         </div>
-                        <input v-else-if="filter.data_type == 'integer'" type="text" inputmode="numeric" pattern="[0-9]*" v-model="filter.selectedValue" class="form-control form-control-sm flex-grow-1 flex-shrink-1">
-                        <input v-else type="text" v-model="filter.selectedValue" class="form-control form-control-sm flex-grow-1 flex-shrink-1">
+                        <input v-else-if="filter.data_type == 'integer'" type="text" inputmode="numeric" pattern="[0-9]*" v-model="filter.selectedValue" class="form-control form-control-sm flex-grow-1 flex-shrink-1" :class="{ 'is-invalid': filter.selectedValue === '' }">
+                        <input v-else type="text" v-model="filter.selectedValue" class="form-control form-control-sm flex-grow-1 flex-shrink-1" :class="{ 'is-invalid': filter.selectedValue === '' }">
```
zarino, line 54: Not sure whether this is what you intended, but it appears this `is-invalid` class is only applied _after_ the input has been edited at least once before. ie: when the input is brand new (and, in most cases, empty), it’s not marked as `is-invalid`, but if you enter some text, then delete it, the input is correctly marked as `is-invalid`.

TBH this actually feels ok to me – always feels nasty immediately displaying an empty input as invalid, before the user’s had a chance to modify it. But I guess it begs the question – why display invalid classes at all, if there are cases where an empty input might sneak through regardless. 🤷

zarino: I guess an alternative (although, with more complex JavaScript) would be to only mark the inputs as invalid _after_ the user has pressed the "Update map" / "Update table" button.

At least then you’d also get some interactive feedback when you press the button and nothing happens. Right now, as it is, if you miss (or misunderstand) the invalid styling, it just feels like the button is broken and not responding to clicks.

zarino (approved):
Generally this seems good to me, and it works locally, so I guess we could merge and deploy.

But I’ve left comments with one suggestion on simplified javascript, and some pondering of the best time to display the `is-invalid` styling on the inputs. I wonder whether you have any thoughts?

hub/templates/hub/explore.html
struan: In theory everything has a default value so we should never show an empty box initially.

zarino: The filter for "MP positions (job titles)" just presents a free text input, and doesn’t have a default value.

It looks like "MP Select Committee memberships" also displays a free text input – but TBH I think that’s a bug, and it should _actually_ be displaying a `<select multiple>` like the "MP APPG memberships" dataset.

struan: oh yes, those. As you say though I think it's fine. FWIW, I think the select committee one is ok text. There's 50 odd committees at the moment which is a fairly unwieldy dropdown.

zarino: Yeah, a free text match on committee name makes sense, but not when it’s prefixed by "is one of" and followed by the ctrl-click/cmd-click help text (both of which should only be displayed next to a `<select multiple>`)…

Screenshot 2023-11-29 at 14 46 55

Anyway, that feels like a separate ticket though.

Issue #333, opened 2023-11-03
Should be able to add "DataType" datasets as shaders on Explore map and additional columns on Explore table


zarino:
As per title. Examples of datasets that don’t currently support shading/columns are: 

- Focaldata/HNH MRP polling
- Defra flood risk
- ONS constituency age distribution
- Ethnicity
- Socio-economic status

We need to make shading by these datasets, and adding them as column columns in table view, work.

I wonder whether this might be an opportunity (our last opportunity?) to reappraise whether the `DataSet`/`DataType` distinction is actually helping?

zarino:
2bc90983b006a549f3c72c7fd5f4699067a4142b enabled _shading_ on these datasets, but we still can’t add them as additional columns in the Explore table view.

- [x] "Add shader" (map view)
- [ ] "Add a column" (table view)

PR #357, opened 2023-11-28
enable shading on data sets with multiple data types

struan:
e.g flood risk

for #333

### DIFF
--- a/hub/static/js/explore.esm.js
+++ b/hub/static/js/explore.esm.js
@@ -192,6 +192,10 @@ const app = createApp({
     addShader(datasetName) {
       this.shader = this.getDataset(datasetName)
 
+      if (!this.shader.selectedType && this.shader.types) {
+        this.shader.selectedType = this.shader.types[0].name
+      }
+
       trackEvent('explore_shader_added', {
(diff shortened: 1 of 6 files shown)


hub/static/js/explore.esm.js
```diff
@@ -493,7 +503,7 @@ const app = createApp({
         case 'filter':
           return dataset.is_filterable
         case 'shader':
-          return dataset.is_shadable
+          return ["party", "constituency_ruc"].includes(dataset.name) || !["text", "json", "date", "profile_id"].includes(dataset.data_type)
```
zarino, line 506: Why hard-code these here, rather than relying on the `is_shadable` flag in the dataset model?

zarino (approved):
This works as far as I can tell!

One question inline about the removal of `.is_shadable` in the javascript – not sure why?

Also, just to flag, #333 also talked about adding DataType datasets as additional columns in the Explore table output, and—while this PR allows me to _select_ DataType datasets in the "Add column" modal—when the table is updated, the response from the server is a CSV with no body rows. Intentional?

PR #358, opened 2023-11-30
DATASET: HFTF trained constituents

### LINKED ISSUE #162: Hope For The Future active/trained constituents
Data is in the shared folder. Currently split into "groups" vs "individuals" (although I’m not sure we need to keep that distinction) and "active" vs "paused" (which, again, might not be _massively_ useful metadata).

alexander-griffen:
Fixes #162

### DIFF
--- a/hub/management/commands/import_hftf_trained_constituents.py
+++ b/hub/management/commands/import_hftf_trained_constituents.py
@@ -1,0 +1,65 @@
+from django.conf import settings
+
+import pandas as pd
+
+from hub.models import AreaData, DataSet
+
+from .base_importers import BaseImportFromDataFrameCommand
+
+
+class Command(BaseImportFromDataFrameCommand):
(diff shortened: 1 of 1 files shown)


struan (approved):
Looks good. I don't really understand the data given we are counting a group as 1 but also "one or more individuals" as the same. Possibly it's just the dataset name needs clarifying though?

zarino:
ACTION: @alexander-griffen to seek advice on the difference between "groups" and "one or more individuals", from our resident HFTF expert @juliacush so we know whether the differentiate them in the dataset, or just treat them as the same thing.

alexander-griffen:
From @juliacush:
> I think group is formalised e.g. Transition Town Exeter, Friends of the Earth Sheffield, and one or more individuals is much more informal, just some people who might be friends/neighbours so the distinction might be meaningless to us!

PR #363, opened 2023-12-04
allow adding data sets in range data types as columns

### LINKED ISSUE #333: Should be able to add "DataType" datasets as shaders on Explore map and additional columns on Explore table

struan:
Fixes #333

### DIFF
--- a/hub/mixins.py
+++ b/hub/mixins.py
@@ -88,6 +88,7 @@ class FilterMixin:
                         "name": dataset.name,
                         "value_col": dataset.value_col,
                         "label": dataset.label,
+                        "header_label": dataset.label,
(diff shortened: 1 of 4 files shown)


zarino (approved):
Looks good to me! And works locally.

PR #365, opened 2023-12-05
auto convert old cons data to new cons data

### LINKED ISSUE #360: New / updated import scripts to translate old constituency datasets to new constituencies

struan:
This adds a script, plus updates all the importers to automatically convert suitable datasets from the old constituency data to the upcoming ones. Essentially it does anything that is a percentage and is evenly distributed over the area or people in the area. It also adds a notice to indicate where a dataset has been automatically converted:

Screenshot 2024-01-11 at 11 39 58

Fixes #360


zarino (commented):
@struan I’ve attempted to tidy up the display of the "disclaimer" in the template, by moving it into the model. My instinct, though, is that the string explaining the conversion should be set _by the script that actually does the conversion_, rather than us having to guess which generation was converted to/from what.

In fact, perhaps `DataType.auto_converted` (a `BooleanField`) should actually be `DataType.auto_converted_from_area_type` (a `ForeignKey` to `AreaType`, defaulting to `None`) so that the view/template can automatically fetch the name and details of the `AreaType` from which the dataset was auto-converted?

struan:
I've made some updates which I think answer most of your issues. I'm not totally sure about the Foreign key thing because I'm not sure what we'd do with it.

PR #375, opened 2023-12-19
DATASET: Christian aid campaign organisers

### LINKED ISSUE #163: Christian Aid campaign organisers
Only a few postcodes here, saved in the shared Drive.

alexander-griffen:
Fixes #163 

Has such few data that the average value for it turns up as '0.0' for now, but this might not be a problem, especially if we get an update with more data.


zarino (approved):
I’ve tweaked the dataset description to actually describe what the campaign organisers are – I _thought_ we’d already done this for all the other datasets, but maybe I’m imagining it. Anyway, regardless, it’s where we should be heading. A "description" that just repeats the dataset label is pointless.

Oh I also spotted that you’d forgotten to change the internal name of the `DataSet` so I’ve fixed that too.

I haven’t actually run the import, so you’ll want to check the data and dataset metadata all comes out as expected, before squashing, merging, deploying, and running on the live site.

struan:
oh, and a small formatting/code fixup

Issue #383, opened 2024-01-12
Public release snagging


zarino:
Issues spotted so far:

- [x] Various dataset issues – https://github.com/mysociety/local-intelligence-hub/issues/364
   - https://github.com/mysociety/local-intelligence-hub/pull/396
- [x] New/old constituency replacement message at top of area page looks bad.
   - https://github.com/mysociety/local-intelligence-hub/pull/384
- [x] Numeric tables on the area page aren’t being converted into bar charts.
   - https://github.com/mysociety/local-intelligence-hub/pull/386
- [x] HOPE not hate MRP dataset cards on area page are missing the bright orange "not to be shared" warning text that used to appear in their card footers.
    - @zarino has confirmed with TCC that they want all members-only data (not just HNH polling) to have a warning like this, so that members know what they can and can’t share publicly.
    - https://github.com/mysociety/local-intelligence-hub/pull/402
- [x] "MP" tab on WMC23 area pages should be renamed "PPCs"
   - https://github.com/mysociety/local-intelligence-hub/pull/385
- [x] We’re missing dataset release/import dates in card footers – didn’t we already complete the work that makes these available? Have we just forgotten to add them to the template?
   - https://github.com/mysociety/local-intelligence-hub/pull/387
- [x] "Favourites" tab on Area page should be hidden for non-logged-in users – @zarino 
   - Part of https://github.com/mysociety/local-intelligence-hub/pull/392
- [x] "Favourite" hearts on Area page (and in Explore page modals – although I think this is already the case) should be hidden for non-logged-in users – @zarino 
   - Part of https://github.com/mysociety/local-intelligence-hub/pull/392
- [x] "TCC member?" advert in Explore page sidebar should _not_ appear for logged-in users
   - Part of https://github.com/mysociety/local-intelligence-hub/pull/392

Bug reports from Bronwen, based on staging site as of 2024-01-17:

- [x] Some datasets ("Support for net zero" and "Estimated households in fuel poverty" at least) have the wrong "Minimum" and "Average" figures when you unfold "About this dataset" on the Explore page. I think it’s counting Northern Irish / Scottish constituencies as "0", when really the "Support for net zero" source data didn’t include NI constituencies, and the "households in fuel poverty" source data didn’t include Scottish or NI constituencies, so they should be excluded from all calculations. Notably, when these datasets are used as a shader, the correct minimum values are shown – @struan 
[...]

PR #385, opened 2024-01-12
Snagging fix: Rename "MP" tab to "PPCs" on WMC23 area pages

zarino:
Part of #383 snagging.

Screenshot 2024-01-12 at 12 52 46


struan (approved):
👍

Issue #377, opened 2024-01-08
Allow users to subscribe to email notifications about new datasets (with optional signup to TCC and mySociety newsletters too)


zarino:
@MyfanwyNixon suggested that we could allow users to sign up to be notified when new datasets are added to the Hub. And while they’re at it, they could also (optionally) sign up to TCC and/or mySociety’s email newsletters. Nice idea!

Question for @MyfanwyNixon – was the idea that the notifications would be sent via a separate Mailchimp list?

Assuming this ticket will make heavy use of Mailchimp, whoever comes to implement it may want to check out [how caps interfaces directly with the Mailchimp API](https://github.com/mysociety/caps/blob/74603adf02d52c1ea3ec9cf6af4046eea119c3f7/caps/views.py#L791-L829) for code we could reuse in LIH.

struan:
relevant: https://mailchimp.com/developer/marketing/api/lists/

zarino:
@struan In case it helps you design the form, I’m imagining something like:

<img width="1432" alt="Screenshot 2024-01-17 at 12 05 14" src="https://github.com/mysociety/local-intelligence-hub/assets/739624/d3f07b33-4fca-4da8-b400-d6d3c8e3fb15">

Where the markup (again, in case that’s useful – expect it’ll be a pain to set up something _exactly_ like this, using Django and `bootstrap-django` forms) for that was:

```html
<form>
    <h2 class="h4">Want notifications when we add new data?</h2>
    <fieldset>
        <legend>Let us know your email address and we’ll notify you about new data:</legend>
        <div class="mb-3">
            <label class="form-label" for="name">Name</label>
            <input type="text" id="name" class="form-control">
        </div>
        <div class="mb-3">
            <label class="form-label" for="email">Email address</label>
            <input type="email" id="email" class="form-control">
        </div>
    </fieldset>
    <fieldset>
        <legend>You can also, optionally, join our newsletters for more inspiration on data, democracy, and climate action:</legend>
        <div class="form-check">
            <input class="form-check-input" type="checkbox" value="" id="check1">
            <label class="form-check-label" for="check1">
                mySociety
                <a href="https://www.mysociety.org/privacy/">(privacy policy)</a>
            </label>
        </div>
        <div class="form-check">
            <input class="form-check-input" type="checkbox" value="" id="check2">
            <label class="form-check-label" for="check2">
                The Climate Coalition
[...]

PR #392, opened 2024-01-16
members only datasets frontend v2

struan:
This does three things:

*  hides non public datasets from unauthenticated users
* updates the area page layout a bit to tidy up the resulting empty boxes etc for unauthenticated users
* turns off the mandatory login for all pages

Fixes https://github.com/mysociety/local-intelligence-hub/issues/280

This replaces #312 which was too out of sync with the area page updates.

### DIFF
--- a/hub/fixtures/area_data.json
+++ b/hub/fixtures/area_data.json
@@ -11,7 +11,8 @@
       "source": "https://parliament.uk",
       "table": "areadata",
       "areas_available": [1],
-      "category": "place"
+      "category": "place",
+      "is_public": "True"
     }
   },
   {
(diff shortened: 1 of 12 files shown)


zarino:
@struan I’ve pushed a fixup commit that adds a yellow section to the bottom of the area page, advertising the extra data for logged in members:

Screenshot 2024-01-17 at 09 49 20

Migrations hell prevented me actually testing this out, so I did the work in `main` and then cherry-picked into this branch, guessing that `{% if not user.is_authenticated %}` was the right thing to check in the template – please replace if not :-)

(FYI longer term, I think we’ll split this yellow section into two columns, and use one of the columns to host the email signup form required by #377.)

PR #394, opened 2024-01-18
Social sharing button on area page

zarino:
This all works and is good to go, but I’ve had to use a CDN version of popper.js, because it turns out the version we’re currently loading doesn’t work (it’s a UMD module being incorrectly loaded as an ESM module) and we just haven’t ever noticed because this Share button dropdown is the first element on the site that actually _uses_ popper.js.

[There’s an ESM version of popper.js here](https://www.jsdelivr.com/package/npm/@popperjs/core?tab=files&path=dist%2Fesm), but it’s a directory of files, and it seems better, to me, to compile them all into a single bundle somehow, to cut down on client-side requests.

I’ve dropped @dracos a line about investigating a solution later this week / next week, as his experience with compiling/bundling/tree-shaking JavaScript is far beyond mine.


struan:
I've got the esbuild branch working although at the cost of I think importing more of bootstrap than is ideal mostly because `import { Modal } from 'bootstrap/js/dist/modal'` does not get you a working Modal :|

I'll have a look at deploying this on the staging site tomorrow I guess.

dracos:
Oh, that's annoying/weird, bootstrap's source modal.js only exports Modal so you'd hope it could work, wonder what's happening there. Does two lines `import 'bootstrap/js/dist/modal'; import { Modal } from 'bootstrap/js/dist/modal'` work at all/different? I don't see why it would but who knows, I've probably misunderstood.

struan:
Sadly, still doesn't work. I am going to leave it as is for now to see if I can get this all in place, and then come back and revisit the bootstrap side of things later.

zarino:
@struan Are we good to merge and deploy this on media launch day?

struan:
@zarino I've rebased it and given it a quick test so, yup, should be good.

PR #396, opened 2024-01-18
WIP: Dataset cleanup

### LINKED ISSUE #364: Datasets need a clean-up
A number of datasets are now _technically_ shadeable (thanks to 2bc90983b006a549f3c72c7fd5f4699067a4142b) but their importers still set `is_shadeable: False`. These should be changed to `is_shadeable: True`:

- [Air quality dataset](https://github.com/mysociety/local-intelligence-hub/blob/30e9da35be5f977ed9002456823a3472952ba9e1/hub/management/commands/import_air_quality_data.py#L34)
- [Age distribution dataset](https://github.com/mysociety/local-intelligence-hub/blob/30e9da35be5f977ed9002456823a3472952ba9e1/hub/management/commands/import_area_age_data.py#L36)
- [Ethnicity dataset](https://github.com/mysociety/local-intelligence-hub/blob/30e9da35be5f977ed9002456823a3472952ba9e1/hub/management/commands/import_ethnicity_data.py#L33)
- [Flood risk dataset](https://github.com/mysociety/local-intelligence-hub/blob/30e9da35be5f977ed9002456823a3472952ba9e1/hub/management/commands/import_flood_risk_data.py#L28)
- [HNH MRP data](https://github.com/mysociety/local-intelligence-hub/blob/30e9da35be5f977ed9002456823a3472952ba9e1/hub/management/commands/import_hnh_polling_data.py#L45)
- [Socio-economic status dataset](https://github.com/mysociety/local-intelligence-hub/blob/30e9da35be5f977ed9002456823a3472952ba9e1/hub/management/commands/import_socio_economic_status_data.py#L31)

I also spotted a couple of datasets from public sources that are missing a `source_url`:

- [Flood risk dataset](https://github.com/mysociety/local-intelligence-hub/blob/30e9da35be5f977ed9002456823a3472952ba9e1/hub/management/commands/import_flood_risk_data.py) – should link to https://www.data.gov.uk/dataset/bad20199-6d39-4aad-8564-26a46778fd94/risk-of-flooding-from-rivers-and-sea
- [Vote share at last election](https://github.com/mysociety/local-intelligence-hub/blob/30e9da35be5f977ed9002456823a3472952ba9e1/hub/management/commands/import_last_election_data.py) – should link to https://commonslibrary.parliament.uk/tag/elections-data/ (better than nothing, covers multiple years)

These ones are also missing a `source_url`, but you can’t currently tell because of the way the area page "groups" together multiple datasets into a single card. Still, once #210 is completed, the missing `source_url` will be more obvious:

- [Trussell Trust foodbank count](https://github.com/mysociety/local-intelligence-hub/blob/b03c57657adbb542776195dac49a8b6b01253568/hub/management/commands/generate_foodbank_count_csv.py) – could link to https://www.trusselltrust.org/get-help/find-a-foodbank/
- [National Trust properties](https://github.com/mysociety/local-intelligence-hub/blob/554d59204e02a3b93dbef3de858a20c2c1bdab3a/hub/management/commands/generate_national_trust_properties_csv.py) – could link to https://www.nationaltrust.org.uk/search
- [WI group locations](https://github.com/mysociety/local-intelligence-hub/blob/ae67f860d0fee99efa15eecab55efdfae9b05112/hub/management/commands/generate_wi_groups_csv.py) – could link to https://www.thewi.org.uk/wis-a-z

The Survation / Renewable UK datasets _have_ a `source_url`, but it’s not the best page to link to. https://www.survation.com/polling-in-every-constituency-in-britain-shows-strong-support-for-building-wind-farms-to-drive-down-consumer-bills/ would be a better `source_url`, since it includes a link to the raw data.

Many datasets are missing values for `unit_type` and `unit_distribution` because their import scripts don’t set them. It looks like [someone wrote a database migration to add these values](https://github.com/mysociety/local-intelligence-hub/blob/c21f2c3af32c47958d2dc6c0f79cd889024069d4/hub/migrations/0051_fill_constituency_change_annotations.py), but forgot to also modify the import scripts (meaning, if you run the migrations _before_ importing the data—as will typically be the case when creating a local dev environment from scratch—the migrations operate on an empty database, and the datasets later get created without the `unit_type` and `unit_distribution` values). Import scripts that were written _after_ [these migrations](https://github.com/mysociety/local-intelligence-hub/blob/c21f2c3af32c47958d2dc6c0f79cd889024069d4/hub/migrations/0051_fill_constituency_change_annotations.py) are also missing the values – such as the [Flood risk dataset](https://github.com/mysociety/local-intelligence-hub/blob/30e9da35be5f977ed9002456823a3472952ba9e1/hub/management/commands/import_flood_risk_data.py#L28).

I also think a number of import scripts neglect to set an `area_type` for their datasets, meaning the datasets don’t appear on constituency pages.

alexander-griffen:
Fixes #364


hub/management/commands/import_brexit_votes.py
```diff
@@ -29,6 +29,8 @@ class Command(BaseImportFromDataFrameCommand):
                 "is_filterable": True,
                 "is_shadable": True,
                 "comparators": DataSet.numerical_comparators(),
+                "unit_type": "raw",
```
struan, line 32: percent, surely?

hub/management/commands/import_christian_aid_group_locations.py
```diff
@@ -47,6 +47,8 @@ class Command(BaseAreaImportCommand):
         "default_value": 0,
         "is_filterable": True,
         "comparators": DataSet.numerical_comparators(),
+        "unit_type": "raw",
+        "unit_distribution": "people_in_area",
```
struan, line 51: I don't think "people_in_area" is correct here. I think "point" is possibly most accurate, but I suspect we possibly need some more categories. There's more of these further down.

hub/management/commands/import_onshore_windfarms.py
```diff
@@ -30,6 +30,8 @@ class Command(BaseConstituencyCountImportCommand):
         "is_filterable": True,
         "is_shadable": True,
         "comparators": DataSet.numerical_comparators(),
+        "unit_type": "raw",
+        "unit_distribution": "physical_area",
```
struan, line 34: I think this is a "point" dataset as they're not even spread.

hub/management/commands/import_ruc_data.py
```diff
@@ -30,6 +30,8 @@ class Command(BaseImportFromDataFrameCommand):
                     dict(title="Urban", shader="grey"),
                 ],
                 "comparators": DataSet.in_comparators(),
+                "unit_type": "percentage",
```
struan, line 33: I think this is raw data as it's just a value, and not a number.

hub/management/commands/import_wildlife_trust_regions.py
```diff
@@ -0,0 +1,43 @@
...
+                "label": "Wildlife Trust Regions",
+                "description": "The Wildlife Trust Region(s) that a constituency falls under.",
+                "data_type": "text",
+                "category": "place",
+                "source_label": "Data from The Widlife Trusts",
+                "source": "https://www.wildlifetrusts.org/",
+                "source_type": "csv",
+                "table": "areadata",
+                "is_filterable": True,
+                "comparators": DataSet.in_comparators(),
```
struan, line 30: missing unit_type and unit_distribution

hub/management/commands/import_wildlife_trust_regions.py
```diff
@@ -0,0 +1,43 @@
+from django.conf import settings
```
struan, line 1: did you mean to include a whole file in here?

struan (changes requested):
There's a chunk of the group count type things that I'm not sure the distribution category is correct. Mostly this looks good though.

hub/management/commands/import_wildlife_trust_regions.py
alexander-griffen: This shouldn't have gotten committed at all, and was a mistake!

hub/management/commands/import_wildlife_trust_regions.py
alexander-griffen: I did not

PR #404, opened 2024-01-22
Split datasets in Explore page modal into categories

### LINKED ISSUE #206: Improve experience of browsing / searching for datasets in “Add filter” modal
The current layout (showing only "featured" datasets, in a single randomly sorted list, with all the other datasets hidden behind a click) isn’t ideal for seeing what data you can filter by.

Testing suggested maybe we let users browse by dataset category (MP, Movement, etc) inside the “Add filter” / “Add shader” modals.

zarino:
Fixes #206.

### DIFF
--- a/hub/static/css/_explore.scss
+++ b/hub/static/css/_explore.scss
@@ -41,6 +41,18 @@ $osm-sea-blue: rgb(168, 212, 243);
     background: $osm-sea-blue;
 }
 
+.explore-modal {
+    // Full-width modal, even on large screens.
+    .modal-dialog {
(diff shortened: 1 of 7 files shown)


struan (approved):
This looks good. Quite wide though so one concern is that if you're looking at this on a tablet in portrait view then you've got a lot of scrolling to get to the place and movement categories. Not sure how much of a use case that is.

I've pushed a fixup to cope with .forEach not returning an array, and also to fix a formatting linting thing.

struan:
Obvs it's even more scrolling on a phone but I am not sure anyone is going to be using this site on a phone

PR #407, opened 2024-01-22
Remove pointless descriptions from MRP datasets

zarino:
Part of dataset cleanup in #364. (I thought about making this commit as part of @alexander-griffen’s clean up of datasets in #396, but figured separate is probably easier to review.)

The new dataset modal in #404 made it clear how repetitious most of the MRP datasets’ descriptions were – simply rewording the dataset label with "estimated percentage of…" at the start.

Hinting that the MRP figures are indeed _estimates_ is probably a nice thing to do, but not if it means as much clutter as this. Let’s revisit again in the near future, especially as we start to get non-technical users on the platform, who might not know what MRP polling even is.

@struan I’m not sure whether this will _remove_ the existing descriptions when the import scripts are next run, or whether we’ll need to do that by hand?

### DIFF
--- a/hub/management/commands/import_cost_of_living_mrp_data.py
+++ b/hub/management/commands/import_cost_of_living_mrp_data.py
@@ -45,18 +45,6 @@ class Command(BaseImportFromDataFrameCommand):
         "mental_health_cost_of_living": "Mental health has been impacted by the cost of living crisis",
     }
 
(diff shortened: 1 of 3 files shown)


struan (approved):
It won't replace the existing text so it will need to be deleted by hand.

PR #411, opened 2024-01-23
Update to use a single JS file build with esbuild

struan:
This is based on @dracos's esbuild branch, with updates to make it work with docker, and add a build script for use in both docker and when deployed.


hub/static/js/all.js
```diff
@@ -1,8 +1,11 @@
-import $ from '../jquery/jquery.esm.js'
-import { Chart, registerables } from '../chartjs/chart.esm.js'
+import $ from 'jquery/dist/jquery.slim'
+//import { Chart, registerables } from '../../../vendor/chartjs/js/chart.esm.js'
```
dracos, line 2: These were only for my history of what I'd changed/testing - don't need to keep these in :) Similar in explore.

hub/static/js/area.esm.js
```diff
@@ -66,6 +66,7 @@ function setUpAreaPage() {
     $('.js-favourite-this-dataset, .js-unfavourite-this-dataset').on('submit', function(e){
         e.preventDefault();
         var $form = $(this);
+        // XXX Needs changing to use fetch() or similar
```
dracos, line 69: This one will need changing though to use the slim build of jquery

hub/static/js/explore.esm.js
```diff
@@ -1,6 +1,16 @@
-import { Modal } from '../bootstrap/bootstrap.esm.min.js'
-import { createApp } from '../vue/vue.esm-browser.prod.js'
-import L from '../leaflet/leaflet-1.9.3.esm.js'
+//import { Modal } from '../../../vendor/bootstrap/js/bootstrap.esm.min.js'
+import { Modal } from 'bootstrap'
+//import { Modal } from 'bootstrap/js/dist/modal'
+import "bootstrap/js/dist/collapse"
+import "bootstrap/js/dist/tab"
```
dracos, line 5: If you're importing all of bootstrap in your Modal import, as currently, don't also import the modules here, or they'll be imported twice, sadly.

dracos (commented):
Couple of tiny comments; see we've got the slimming down of chart and jquery (and maybe leaflet?), understand bootstrap didn't work (see my little help there for now at least), but sad to see vue didn't either. My guess is it might be: "You are using the runtime-only build of Vue where the template compiler is not available. Either pre-compile the templates into render functions, or use the compiler-included build." - so something to consider there if that would cut it down considerably.

Issue #341, opened 2023-11-08
Update search to handle new constituencies


struan:
Needs to default to one set to avoid confusion.

New constituencies are in mapit so we can use that look up both types.

Will need to handle choosing if you want new or old constituencies. Options are:
* Pick one to show when using postcode
* Always show disambiguation page for postcode
* Have a choice field next to the search box

This will also need to handle searching for PPCs which means multiple matches per constituency.

PR #412, opened 2024-01-23
only search for MPs

### LINKED ISSUE #410: 500 error when searching by PPC name
Searching for a few high-profile MPs (such as Jeremy Hunt, Oliver Dowden) via the search box on the homepage currently causes a 500 error:

```
Exception Type: MultipleObjectsReturned at /location/
Exception Value: get() returned more than one Person -- it returned 2!
Raised during: hub.views.area.AreaSearchView
```

I _think_ what’s going on is this:

`AreaSearchView` currently calls `Person.objects.get(area=area)`. That `area` comes from a list of objects queried a few lines further up:

```python
areas_raw = Area.objects.filter(
    name__icontains=search, area_type__code="WMC"
)
```

This, on its own, works fine, since WMCs only have a single `Person` – their MP.

But when you search for a person’s name (rather than a constituency’s name), that person’s "area" gets added to the `areas` list.

```python
people_raw = Person.objects.filter(name__icontains=search)

areas = list(areas_raw)
for person in people_raw:
    areas.append(person.area)
```

If that person happens to be standing as a PPC in a `WMC23` area, then that `areas` list now contains a `WMC23` area.

And—you can see where this is going—if that area has multiple PPCs, then `Person.objects.get(area=area)` returns a list of multiple `Person` objects, causing the exception.

struan:
Partly as this avoids issues with matching multiple constituencies and also to avoid multiple person objects being returned for an area.

Fixes #410

### DIFF
--- a/hub/views/area.py
+++ b/hub/views/area.py
@@ -356,7 +356,7 @@ class AreaSearchView(TemplateView):
             areas_raw = Area.objects.filter(
                 name__icontains=search, area_type__code="WMC"
             )
-            people_raw = Person.objects.filter(name__icontains=search)
+            people_raw = Person.objects.filter(person_type="MP", name__icontains=search)
 
             areas = list(areas_raw)
             for person in people_raw:


zarino (approved):
Seems fine. We should make a note to revisit this as part of #341 (where the plan is to switch the search over, from searching MPs and current constituencies, to PPCs and future constituencies, closer to the election).

PR #424, opened 2024-01-25
Add a view that serves up area geometry with cache headers

### LINKED ISSUE #421: work out if we can cause map geometry to be cached
This is a 2ish Mb download and it doesn't change so it should be possible for this to be cached on the browser side which would make the map page start up quicker.

struan:
Do this as a separate view in case we ever have boundaries that do change.

Fixes #421


zarino (approved):
:+1:

PR #425, opened 2024-01-25
Update (or remove) descriptions of datasets

alexander-griffen:
Part of #383 - already updated on production in Django Admin


hub/management/commands/import_wi_group_locations.py
```diff
@@ -37,7 +36,7 @@ class Command(BaseAreaImportCommand):
 
     count_defaults = {
         "label": "Number of Women’s Institute groups",
-        "description": "Number of Women's Institute Groups",
+        "description": "",
```
struan, line 39: is there a reason to make this one a "" and for the others to be removed?

struan (approved):
👍

PR #429, opened 2024-01-29
in progress mailing list signup

### LINKED ISSUE #377: Allow users to subscribe to email notifications about new datasets (with optional signup to TCC and mySociety newsletters too)

struan:
Fixes #377


zarino (approved):
This looks fine to me, but you should check with @MyfanwyNixon that it adds the right interest groups / tags (assuming the changes @ajparsons suggested in https://github.com/mysociety/orgsites/issues/1353 are actually happening).

ajparsons:
They have - it looks new climate one is: group[11745][68719476736]

data is: group[11745][256]

PR #433, opened 2024-02-01
update to point significant votes to twfy voting page

struan:
Slightly short term way to do this.


zarino (approved):
Agree that this is a bit short-term, but it’s good enough for now. If/when there’s a third relevant vote, we can reconsider.

ajparsons:
Just to quickly note here that TWFY does accept the ID as a shortcut, and commons votes has all the needed parts to construct it:

https://www.theyworkforyou.com/divisions/pw-2020-12-02-175-commons

Issue #382, opened 2024-01-11
decide on if/how to do cross promition with TWFY


struan:
Currently LIH has links to a MP pages on TWFY. Do we need anything more?

And do we want to put links on TWFY to LIH. TBH I' am not sure it super makes sense to put it on TWFY as it's not really information about an MP so it feels a bit odd.

ajparsons:
Ok, so meta point here is I want better ability to treat theyworkforyou as a hub to our other services (where that makes sense), and to links on official parliamentary profiles (https://github.com/mysociety/theyworkforyou/issues/1714) etc.

We have *some support for this*, but it's a bit inconsistent:

https://www.theyworkforyou.com/mp/10001/diane_abbott/hackney_north_and_stoke_newington

<img width="629" alt="image" src="https://github.com/mysociety/local-intelligence-hub/assets/8157058/92ffee2b-3b58-4a21-b190-9185c993390f">

Here we have some social links, and the write button which links to WriteToThem. 

I don't think it'd be inappropriate to have a 'Constituency Info' button that goes to LIH. But more generally, It'd be nice to get a little bit of your time documenting how this is currently working - and some notes documenting (or just doing a change) an adaption to the social links feature. Are those buttons the right way to go? Or do we need a little links panel in the profile rather than the top bar?

Relates as bit to https://github.com/mysociety/theyworkforyou/issues/1736

struan:
So, on the TWFY side I think the best option is to populate the [constituency-links.xml](https://github.com/mysociety/parlparse/blob/master/members/constituency-links.xml) with entries for the related LIH pages. This should be a one time thing. We can then update TWFY to use that data on the members page. I suggest we create a ticket on the TWFY side to cover this.

PR #435, opened 2024-02-05
TheyWorkForYou and WriteToThem links in MP section on Area page

zarino:
Adds a new "Write to your MP about this data" link above MP section, and adds another link to TWFY voting record (this time, near to where we’re already displaying votes, which is a bit more intuitive).

Related to #382.

Screenshot 2024-02-05 at 10-53-43 Liverpool Riverside


struan (approved):
I do wonder about finding a better link for the election results than an API results page, although good to link directly to the data source. However, I wonder if adding a link to this page would also be good: https://electionresults.parliament.uk/election/2019-12-12/Results/Location/Constituency/Bermondsey%20and%20Old%20Southwark

PR #440, opened 2024-02-05
Refactor group list/count importers plus import TCC open letter signatory data

### LINKED ISSUE #430: Show which groups/organisations in this constituency have signed TCC’s open letter
TCC are collecting signatures from community groups on an open letter to party leaders:

https://www.theclimatecoalition.org/community-open-letter

Signatories fill in a form that includes a field on "what geographical area their group covers", but that has to be manually gecoded to one or more constituencies by the team at TCC. Once they’ve done this, they provide the constituency name as a column in a spreadsheet.

We need to:

- Import organisation names from that spreadsheet, and display them on the area page, and in the Explore page filters
- Check whether TCC are also "geocoding" to future constituencies (as well as current constituencies)

struan:
This factors out the common code for the various importers that create both a count of groups/locations in a constituency plus a json field showing the names/details of those groups. It then uses that to import the details of signatories to the TCC open letter.

Fixes #430


zarino (approved):
I’ve added some commits here, which go beyond the scope of #430, but felt necessary (like adding source links to dataset cards—I thought we already had them!—adding descriptions, and removing the doubled-up headings in group-type cards).

I wonder, when we’re importing the signatories, whether we should pre-populate a URL for each one, which is just `https://www.google.com/search?s=[signatory name]` so that the items in the tag cloud are, at least, clickable?

PR #443, opened 2024-02-07
WIP: Remove unapplicable datatypes from filter and shader dropdowns

### LINKED ISSUE #441: Survey options appear to be duplicated in the pulldown menu
Steps to replicate: 

From https://www.localintelligencehub.com/explore/ (not logged in), select "Loss of nature and destruction of the environment should be one of the most important concerns for the country" as a dataset and click the pulldown - options seem to be shown twice. 
Screenshot 2024-02-05 at 18 03 02

### LINKED ISSUE #442: Filter and Shader UI should only show DataTypes that match the current selected area_type
When a dataset `.is_range` and _also_ covers multiple `area_type`s (the HNH MRP polling is an example of this, eg: "loss of nature…") then the data type `<select>` boxes include all datatypes, rather than just the ones that match the currently selected `area_type`:

Screenshot 2024-02-07 at 10 15 09

`ExploreDatasetsJSON` needs to expose the `area_type.code` of each datatype in each dataset, and then the vue.js template needs to filter out datatypes that don’t match `this.area_type`.

zarino:
Fixes #442. Fixes #441.

I’m having a bit of an issue testing this out locally, as it seems all of my local `is_range` datasets aren’t showing up in the dataset picker modal for some reason… 🤔 But I _think_ this should work.

### DIFF
--- a/hub/views/explore.py
+++ b/hub/views/explore.py
@@ -114,8 +114,14 @@ class ExploreDatasetsJSON(TemplateView):
                 ds = {**ds, **type_map[d.name]}
             if d.is_range:
                 ds["types"] = [
-                    {"name": dt.name, "title": dt.label}
-                    for dt in DataType.objects.filter(data_set=d)
+                    {
+                        "name": dt.name,
+                        "title": dt.label,
+                        "area_type": dt.area_type.code if dt.area_type else None,
+                    }
+                    for dt in DataType.objects.filter(data_set=d).select_related(
+                        "area_type"
+                    )
                 ]
             datasets.append(ds)
 
--- a/hub/static/js/explore.esm.js
+++ b/hub/static/js/explore.esm.js
@@ -586,6 +586,9 @@ const app = createApp({
         default:
           return true
       }
+    },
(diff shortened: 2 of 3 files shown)


hub/views/explore.py
```diff
@@ -114,7 +114,11 @@ def render_to_response(self, context, **response_kwargs):
                 ds = {**ds, **type_map[d.name]}
             if d.is_range:
                 ds["types"] = [
-                    {"name": dt.name, "title": dt.label}
+                    {
+                        "name": dt.name,
+                        "title": dt.label,
+                        "area_type": dt.area_type.code if dt.area_type else None,
+                    }
                     for dt in DataType.objects.filter(data_set=d)
```
struan, line 122: this should have a `.select_related("area_type")` otherwise the `dt.area_type.code` will be generating a SELECT for each loop entry.

struan (approved):
Maybe either split the styling changes in to a separate commit or at least comment in the commit what's happening there? Plus one small bit of SQL efficiency, otherwise good.

PR #448, opened 2024-02-08
Refine vote title/disclaimer

ajparsons:
Minor tweak to wording to reflect that this is a subset (usefully chosen!) rather than comprehensive.


zarino (commented):
@ajparsons Do you think maybe we should spell out "Early Day Motions"? "EDMs" feels a bit jargony to me.

PS, no need to touch `_mp_data.html` – it’s no longer used, and will be removed by #449.

struan (approved):

PR #450, opened 2024-02-08
update mp job titles import script

struan:
Imports MP job titles and displays them on the area page in the MP box, which I'm not totally convinced by.

Also, this includes Select Committee Chairs as I assume that's useful information.

NB: good examples of extremes for multiple jobs/length of job title are Angela Rayner and Ben Lake.

Screenshot 2024-02-12 at 12 28 47

Screenshot 2024-02-12 at 12 28 37

Screenshot 2024-02-12 at 12 29 50


zarino:
@struan I wonder, would it look better if we did something like this instead?

Screenshot 2024-02-12 at 12 43 14

struan:
Putting them there works, not sure about using the tag thing though as some of the titles are very long and it cuts them off:

Screenshot 2024-02-12 at 13 43 06

Screenshot 2024-02-12 at 13 44 32

If I do it as before though I get:

Screenshot 2024-02-12 at 14 08 13

Screenshot 2024-02-12 at 14 08 28

zarino:
Oh wow, that Ben Lake position text is unbelievable!

Yeah, I agree, tags aren’t the right thing here – these titles aren’t clickable, for a start (which is where the tags work best), and they’re not short either (so they get cut off, as you said).

I think just using plain text is fine, as long as we can differentiate the different positions from one another – when they’re only separated by a linebreak, it’s quite difficult to parse where one position ends and the next begins. I’ll try fiddling with font-size and spacing between positions 👍

zarino:
Ok @struan I’ve tweaked this so that the positions are structured in a list, which lets us space them out a bit. I’ve also dropped the subheading to save a bit of space (and prevent the repetition of "Positions" in one heading right above the other), and changed the vague "collated by" attribution.

I’m sure there will be some MPs where this layout looks odd, but I guess we can keep an eye out for them later.

zarino (approved):

PR #452, opened 2024-02-12
Handle missing MP on area page

### LINKED ISSUE #393: Handle current constituencies that do not (currently) have an MP
e.g currently Wellingborough does not have an MP due to a recall petition but LIH shows:

Screenshot 2024-02-12 at 12 47 34

I've updated the mp election result script to set the end_date of an MP so we just need to update the page to do something with that. I suggest we possibly continue to show the MP but add some indication that they are no longer sitting. The parliament API does have end reasons so possibly we can use that.

At some point we will need better handling of this to account for changes in MP, although that is possibly not relevant here.

struan:
Fixes #393 

Screenshot 2024-02-12 at 16 21 55

### DIFF
--- a/hub/admin.py
+++ b/hub/admin.py
@@ -165,5 +165,5 @@ class PersonAdmin(admin.ModelAdmin):
     )
     search_fields = (
         "name",
-        "area",
+        "area__name",
     )
--- a/hub/templates/hub/area.html
+++ b/hub/templates/hub/area.html
@@ -114,8 +114,18 @@
 
         <section class="mb-5 mb-lg-6 area-section" id="mp">
             {% if area_type == "WMC" %}
+            {% if no_mp %}
+              <h2 class="text-primary">This constituency has no MP right now</h2>
+              <p style="max-width: 37em;" class="mb-4">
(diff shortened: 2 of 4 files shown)


zarino (approved):
I’ve added a fixup commit that changes the area page markup a bit. It should look something like this, but with less missing data:

Screenshot 2024-02-12 at 17 53 13

I’m guessing the missing data is just my local database being out of date, rather than this being what users would actually see for a past MP. Could you check it at your end, @struan? The aim is that we’d display the previous MP’s name, (party, ideally), majority, and end date, as "featured" cards; with election date and identifiers/socials as non-featured cards; and memberships/job_titles/votes removed entirely.

struan:
I'm not sure the "there is a by-election scheduled" thing will always be true but, meh, fine.

zarino:
> I'm not sure the "there is a by-election scheduled" thing will always be true but, meh, fine.

Yeah, I know, I felt a bit cheeky putting that in. I wonder whether maybe `<a href="…">There may be a by-election scheduled for this constituency.</a> …` would be better?

struan:
I'll fix that before deploying it

Issue #406, opened 2024-01-22
Improve print layout of Area page


zarino:
Area page currently prints out in mobile view. At the very least, we should keep the grid layout, so that what you get on the printed page resembles what you’re seeing on screen.

PR #454, opened 2024-02-13
Print stylesheets

lucascumsille:
Fixes: https://github.com/mysociety/local-intelligence-hub/issues/406

-Reduced the base font size to maintain readability without making the information on the page appear too cramped. 
-Added classes to HTML so in the printed version only element that are necessary are displayed.

IMPORTANT: It needs testing with more data.
Here is a preview of how it looks:
[Henley.pdf](https://github.com/mysociety/local-intelligence-hub/files/14265251/Henley.pdf)

### DIFF
--- a/hub/static/css/_print.scss
+++ b/hub/static/css/_print.scss
@@ -1,0 +1,40 @@
+@media print {
+  html {
+      font-size: 14px;
+      background-color: $white;
+  }
+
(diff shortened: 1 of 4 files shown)


zarino (approved):
I’ve tested this out with more data in my database, and it looks great! :+1:

PR #456, opened 2024-02-13
handle constituencies with no MP in table and map

struan:
if the number of MPs listed isn't the same as the number of areas then fill in the missing ones with a no current mp label so the constituencies still appear in the list, rather than as previously with the old MP


zarino (approved):
Looks good to me :+1:

PR #458, opened 2024-02-13
add a sitemap

### LINKED ISSUE #371: Generate a sitemap containing all pages of the site
The area pages, in particular, are only accessible via search and client-side javascript, so we should include those in a sitemap, for crawlers / search engines to index.

I wonder whether https://docs.djangoproject.com/en/4.2/ref/contrib/sitemaps/ might help?

Google’s guidance on sitemaps: https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap

struan:
Fixes #371

### DIFF
--- a/hub/sitemap.py
+++ b/hub/sitemap.py
@@ -1,0 +1,31 @@
+from django.contrib.sitemaps import Sitemap
+from django.urls import reverse
+
+from hub.models import Area
+
+
+class StaticViewSitemap(Sitemap):
+    def items(self):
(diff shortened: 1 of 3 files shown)


zarino (approved):
Looks good to me :+1: I’ve added a fixup with the `future-constituencies` static page, because why not.

We’ll need to remember to check Google Search Console a little while after this is deployed, to make sure Google’s picked up the new sitemap.

PR #462, opened 2024-02-14
Further JS and CSS improvements for page load increases

### LINKED ISSUE #397: consider splitting up JS a bit so we're not loading things we don't need on the front page
At the moment all the pages use `all.js` which includes both the chart js and the explore js so we pull in a load of files we don't really need at this point. We might want to split those out at some point.

The counter argument to this is that anyone using the site is presumably going to need those and they are cached for a long time so might as well load them now and the later pages will be fine, especially as, I think, those are non blocking downloads so the page will render anyway.

struan:
This modularises the JS futher, adds in some bootstrap dist files so we can only load the bits of that we don't need.

It also makes the google tag manager loading async which should stop some blocking.

Finally it creates a front page specific CSS file to cut down on loading that.

More work could probably be done with the CSS on some of the other static pages.

Fixes #397


zarino (approved):
I’ve added a commit that swaps `map.png` for a new `map.webp` which halves the filesize for no perceptible difference. And I spotted some vertical space was missing between the "filters" on the homepage, so added that back in.

You commit mentioned removing te xxl breakpoint on the homepage – not sure whether it’s related, but I am seeing that `.container` elements on the homepage now fill the viewport whenever it’s narrower than ~1300px (rather than locking to a series of pre-determined widths). This isn’t a big problem, but it does mean that, if you navigate from the homepage to another static page (ie: anything but `/explore`) then things like the header and footer jump to narrower widths. Not a big deal, but it does make me wonder whether there’s a way to keep the `.container`s working, without adding much more bloat to the homepage css…?

PR #467, opened 2024-02-15
improve MP import script

struan:
This uses the TWFY MP CSV as the base information rather than wikidata as it doesn't seem to keep up to date, and also this way the data is in our control which is probably good.

It does mean we need to get three sets of data to enable all the same information but again, the MP list and the parliament ID, which is required for election results among other things, is now using our data which is probably better.

### DIFF
--- a/hub/management/commands/import_mps.py
+++ b/hub/management/commands/import_mps.py
@@ -10,6 +10,7 @@ from django.db.models import Count
 from django.db.utils import DataError
 
 import magic
+import pandas as pd
 import requests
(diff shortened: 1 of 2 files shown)


zarino (approved):
Tested this locally and it worked. With @ajparsons’ help I added a few lines to temporarily work around the lack of `parlid`s for the newly elected MPs for Wellingborough and Kingswood. I’ll create a new ticket to remind us to fix that up later – perhaps before the next byelection at the end of Feb.

PR #471, opened 2024-02-19
Add canonical link to page head

zarino:
Google Search Console notified us about a `Duplicate without user-selected canonical` error when crawling `https://www.localintelligencehub.com/?utm_source=theyworkforyou&utm_medium=website&utm_campaign=banner`.

Doing a bit of digging, it sounds like [a `canonical` link](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls) is the way Google expect you to tell its crawler what the "official" URI for the page might be, ignoring whatever query params might happen to have led it there.

I use `request.path` rather than `request.build_absolute_uri`, because the latter would _include_ query params, and we don’t want that. (We did [almost the exact same thing](https://github.com/mysociety/caps/pull/606/files) for the Scorecards at the end of last year.)

This _will_ mean that the `/explore` page is always served up with a canonical tag that lacks query params – even when the page has been requested with filters/shaders in the query string – but I think that’s fine?


struan (approved):
In an ideal world we should probably not be using the request to generate this but as a short term fix I think it's fine.

PR #475, opened 2024-02-20
call the Green Party the Green Party everywhere

### LINKED ISSUE #473: fix green party title in Scotland
The last election result importer uses Green Party of England and Wales in Scotland when it should use Scottish Green Party.

struan:
Standardise on this rather than use the correct names for the Scottish and England/Wales parties. It's not strictly correct but makes it easier for things like "show all constituencies with a green party candidate".

Fixes #473


zarino (approved):
Seems ok, as a quick fix.

Do we need to do the same thing for the [Irish Green Party](https://en.wikipedia.org/wiki/Green_Party_(Ireland))? (There are no current Irish Green MPs, but I wonder whether the party _might_ show up in the "second placed party" and vote share datasets?)

PR #485, opened 2024-02-25
Add Threads option and event tracking to area page social sharing dropdown

zarino:
A user left feedback today:

> Wanted to share the search option to Threads but no option to do so. Its a much better places than Twixter… 

They might have been asking to share a page _other_ than the area page, which is interesting. Sadly they didn’t leave an email address, so we can’t ask.

But as a first step, I’ve added a Threads option to the social sharing dropdown on the Area page.

And to help us get a feel for how much people use these sharing buttons (and whether we might want to add them to other pages) I’ve added Google Analytics click tracking. Handily, because all the social sharing links open in a `_blank` window, I don’t need to prevent page navigation until the event has tracked, or anything – the call to Google Analytics seems to continue in the background, while the social website is opening in a new, foreground tab.

I‘ve added a new `service` event parameter to our Google Analytics setup, so it’ll appear in reports. `page_location` is a built-in event parameter, so appears already.


struan (approved):
@ 👍

PR #489, opened 2024-02-27
DATASET: Wildlife trusts regions

### LINKED ISSUE #317: Import data on the Wildlife Trusts covering each constituency
Wildlife Trusts publish their Trust boundaries as open data:

 https://geoportal-wildlifetrusts.hub.arcgis.com/datasets/wildlifetrusts::wildlife-trust-regional-boundaries-open-data/explore?location=55.083629%2C-3.313256%2C7.46

Would be great to use this to display the names of the Trusts covering a given constituency, on the constituency page. (Note: there are ~48 Trusts, so around a dozen constituencies per Trust, and constituencies might straddle Trust boundaries.)

Extra marks if we can also _filter_ by it on the Explore page with a `<select multiple>`, like we do for political parties.

alexander-griffen:
Fixes #317 

NOTE: The dataset csv can be found in the shared drive.

Additionally, the following script was used to produce it: 
```
import geopandas as gpd

wildlife_trust_regions = gpd.read_file(
    "Wildlife_Trust_Regional_Boundaries_-_Open_Data.zip"
)
constituencies = gpd.read_file(
    "WPC_Dec_2019_Boundaries_UK_BFC_V2_2022_-7467643007870890518.zip"
)


constituencies = constituencies.to_crs("EPSG:4326")
combined = wildlife_trust_regions.overlay(constituencies, how="intersection")

combined["area"] = combined.geometry.area
combined = combined.query("area >= 0.00125")

combined[["Trust", "PCON19NM"]].to_csv("wildlife_trust_regions_in_constituencies.csv")
```

The number used - 0.00125 - was callibrated by incrementing and de-incrementing it until the dataset covered every constituency, but didn't overlap egregiously.

### DIFF
--- a/hub/management/commands/import_wildlife_trusts_regions.py
+++ b/hub/management/commands/import_wildlife_trusts_regions.py
@@ -1,0 +1,53 @@
+from django.conf import settings
+
+import pandas as pd
+
+from hub.models import DataSet
+
+from .base_importers import BaseConstituencyGroupListImportCommand
+
+
+class Command(BaseConstituencyGroupListImportCommand):
+    help = "Import Wildlife trust reserves"
+
+    data_file = (
+        settings.BASE_DIR / "data" / "wildlife_trust_regions_in_constituencies.csv"
+    )
+    cons_row = "PCON19NM"
+    uses_gss = False
+    message = "Importing constituency Wildlife Trust regions data"
+
+    data_sets = {
+        "wildlife_trusts_regions": {
+            "defaults": {
+                "label": "Wildlife Trust Regions",
+                "description": "The Wildlife Trust Region(s) that a constituency falls under.",
+                "data_type": "json",
+                "category": "place",
+                "subcategory": "groups",
+                "source_label": "Data from The Widlife Trusts",
+                "source": "https://www.wildlifetrusts.org/",
+                "source_type": "csv",
+                "table": "areadata",
+                "is_filterable": False,
+                "is_shadable": False,
+                "comparators": DataSet.in_comparators(),
+            },
+            "col": "Trust",
+        }
+    }
+    group_data_type = "wildlife_trusts_regions"
+
+    def get_df(self):
+        return pd.read_csv(self.data_file, usecols=["Trust", "PCON19NM"]).rename(
+            columns={"PCON19NM": "constituency"}
+        )
+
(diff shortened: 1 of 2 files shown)


hub/management/commands/import_wildlife_trusts_regions.py
```diff
@@ -0,0 +1,45 @@
...
+                "label": "Wildlife Trust Regions",
+                "description": "The Wildlife Trust Region(s) that a constituency falls under.",
+                "data_type": "text",
+                "category": "place",
+                "source_label": "Data from The Widlife Trusts",
+                "source": "https://www.wildlifetrusts.org/",
+                "source_type": "csv",
+                "table": "areadata",
+                "is_filterable": True,
+                "comparators": DataSet.in_comparators(),
```
struan, line 32: Not sure this is the best comparator as you need to know the exact name of the trust, unless we populated the options with all the names of the trusts but 40 is possibly unwieldy for this. Might make more sense to make this a text comparator, although then can't look for more than one trust then. Maybe 40 is not that bad, although also, see other comment :|

hub/management/commands/import_wildlife_trusts_regions.py
```diff
@@ -0,0 +1,45 @@
...
+    cons_row = "PCON19NM"
+    uses_gss = False
+    message = "Importing constituency Wildlife Trust regions data"
+
+    data_sets = {
+        "wildlife_trust_regions": {
+            "defaults": {
+                "label": "Wildlife Trust Regions",
+                "description": "The Wildlife Trust Region(s) that a constituency falls under.",
+                "data_type": "text",
```
struan, line 25: This is causing a problem as it means a constituency can only be in one trust which ends up being the last one with that constituency in the sheet. It might make more sense for this to work like groups, although then not filterable although we should probably create a ticket to allow filtering on group names as you can do that in postgres I think.

struan (changes requested):
I think need to change the type of this so we can represent trusts that cover multiple constituencies.

struan (approved):
only small thing is probably want to make the not filterable as no sensible way to do it with json at the moment.

zarino:
@alexander-griffen are you still working on this, or is it ready to merge, deploy, and run? If the latter, do you want to go ahead and do it?

PR #506, opened 2024-03-13
Display local council boundaries and add some basic council data

struan:
This adds local council boundaries to the map along with imports for the population and the type so there is some data there. It also makes minimal changes to the area page for local councils.


hub/management/commands/import_areas.py
```diff
@@ -11,46 +11,71 @@
...
+                name=b_type["name"],
+                code=b_type["code"],
+                area_type=b_type["area_type"],
+                description=b_type["description"],
             )
 
-            a.geometry = geom
-            a.save()
+            if not quiet:
+                print("Importing Areas")
```
zarino, line 55: Would be nice if this said which _type_ of area it’s importing.

hub/management/commands/import_council_data.py
```diff
@@ -0,0 +1,51 @@
...
+
+    area_types = ["STC", "DIS"]
+
+    defaults = {
+        "data_type": "integer",
+        "category": "place",
+        "subcategory": "",
+        "release_date": "February 2023",
+        "label": "Council population",
+        "source_label": "Data from mySociety.",
```
zarino, line 23: Is this the "estimated 2020 population" from https://pages.mysociety.org/uk_local_authority_names_and_codes/datasets/uk_la_future/latest ? Or something else? Would be nice to link to the _actual_ source of the data, rather than our homepage, generically.

zarino: If this _is_ indeed the `pop-2020`, then the `"release_date": "February 2023"` above should change too.

hub/management/commands/import_council_data.py
```diff
@@ -0,0 +1,51 @@
...
+        "source": "https://www.mysociety.org/",
+        "source_type": "csv",
+        "table": "areadata",
+        "default_value": 1000,
```
zarino, line 27: Don’t we typically set this to something close to the average or the mode?

hub/templates/hub/area.html
```diff
@@ -120,9 +120,11 @@ <h3 class="h5"><a href="{% url 'area' area_type=overlap_constituency.old_area.ar
                 </a>
             </li>
           {% endif %}
+            {% if is_westminster_cons %}
```
zarino, line 123: Would be nice if this was two-space indented, like the conditional above ;-)

(I try to always do this inside Django templates, as per the "suggestion" in https://pages.mysociety.org/coding-standards.html)

zarino (commented):
Yep, Area page and Explore page both work for councils :+1:

Have suggested a few improvements.

hub/management/commands/import_council_data.py
struan: we do use the average which is calculated later in the process. This is just a placeholder to fall back on.

zarino (approved):
Overall looks good :+1: Two small things:

There’s too much vertical space above the public opinion section, on council area pages, because the (empty) MP section is still there in the markup. Any way we could skip printing that section entirely, if we’re on a council page?

Screenshot 2024-04-08 at 13 58 18<br>

And is there any way we can get a filter in the DataSet admin, to filter by council type(s)?

Screenshot 2024-04-08 at 13 58 33

PR #519, opened 2024-03-27
Add a fake dataset for the constituency URL

zarino:
Handy when you’re downloading constituency data as a CSV, and want to be able to quickly link back to the Area’s webpage.


struan (approved):
I've pushed a fix for the tests and the URL thing, but otherwise looks good.

PR #520, opened 2024-03-28
Update air quality data

### LINKED ISSUE #180: Air quality / pollution data

alexander-griffen:
Fixes https://github.com/mysociety/local-intelligence-hub/issues/180

Update air quality data to use more recent data and also include WHO standards

This currently does not show up in the UI. Additionally, the following tasks need to be completed on this work:

- Finding a way to show the WHO standards in the UI
- Possibly finding more WHO standards, as I only could find them for some of the pollutants

Screenshot 2024-04-01 at 17 00 55

Screenshot 2024-04-01 at 17 01 33

### DIFF
--- a/hub/management/commands/import_air_quality_data.py
+++ b/hub/management/commands/import_air_quality_data.py
@@ -25,7 +25,7 @@ class Command(BaseCommand):
         "data_type": "float",
         "category": "place",
         "source_label": "Data from Defra.",
-        "release_date": "2021",
+        "release_date": "2022",
         "is_range": True,
         "source": source_url,
         "source_type": "csv",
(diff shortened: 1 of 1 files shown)


zarino (changes requested):
We have a choice here – do we want to just get out a quick change that updates from 2021 to 2022 Defra pollutant data, or do we want to incorporate WHO guidelines?

If this PR is just about updating from 2021 to 2022 Defra data, then:

- We should remove the WHO_AQG stuff.
- We should fix the display of this data on the constituency page – the figures should be rounded floats, not percentages, and should probably just be a table, rather than a bar chart, because comparing bars for totally different pollutants on different scales makes no sense.

If this PR is _also_ about making the air quality data easier to understand, by comparing with WHO guidelines, I’m thinking we should:

- EITHER create a _separate_ dataset that presents comparative figures (ratio? percentage?) for just the three pollutants we have comparable data for.
- OR keep it all in one dataset, but create a custom set of comparators for this dataset, so that users can select options like "is above WHO guidelines" and "is below WHO guidelines", in addition to the normal "is equal to or greater than" / "is less than" options. (This feels like _more_ work to me, and still not as flexible as the separate dataset option.)

I’ve added more detail on this to the original ticket at https://github.com/mysociety/local-intelligence-hub/issues/180

PR #521, opened 2024-04-01
decrease map stroke width to 1 for better legibility

struan:
especially helpful when zoomed out and where there are lots of small areas


zarino (approved):
:+1: I’ve amended your commit to include a corresponding change further down the file, in the `mouseout` handler, to set the stroke weight back to 1 after hovering.

Issue #505, opened 2024-03-11
UI for switching area type on Explore page


zarino:
Design task, broken out of #487.

We currently have a select box, and some accompanying text, on [the Explore page](https://www.localintelligencehub.com/explore/), that enables you to switch between current constituencies and future constituencies:

<img width="257" alt="Screenshot 2024-03-11 at 10 58 28" src="https://github.com/mysociety/local-intelligence-hub/assets/739624/ade7172c-81c6-4b6e-94a7-e149e13c73ce"><br>

We’ll soon need a way to switch between two "levels" of local authority, in addition to these constituencies (so, a total of four options). And eventually there may be many more (eg: Welsh Senedd constituencies, Scottish Parliamentary constituencies, etc).

How should this look and work?

PR #524, opened 2024-04-09
Added UI for area type selection

lucascumsille:
Fixes: https://github.com/mysociety/local-intelligence-hub/issues/505

Added UI for area type in Map page
-"select button" display the area name + description.
-"select button" Opens a modal displaying the area option with an description for each one of them.
-Button content is updated when user selects an option.

Preview here:

https://github.com/mysociety/local-intelligence-hub/assets/13790153/62937753-829f-4f09-a8b4-31583b076787


zarino (approved):
Looks good. FWIW, I’ve added a fixup commit that:

- Improves the wording of the area type descriptions (making them plural, and simplifying the council ones).
- Uses the currently selected area type in the “Filter…” and “Shade…” headers.
   - And while I was doing this, I replaced the `area_header_label` Vue variable (which we were using, previously, to get a short word for the selected area type—like “councils” or “constituencies”—and replaced it with a `short_label` attribute on the area type object, which we can pull out like the area type `label`s and `description`s. Feels more consistent.
- Splits out the previous max-width/margin styling and list-group padding from `.explore-modal`, so that we could apply the padding to both types of modal (now `.dataset-modal` and `.area-type-modal`) while keeping the full-width behaviour just for the `.dataset-modal`.
- Vertically centres the area type modal, using the `modal-dialog-centered` class.
- Adds a `shadow-lg` class to the area type modal, to match the dataset modal.
- Re-indents `area-type-modal.html` and removes the unnecessary `<template>` wrapper.
- Improves vertical spacing and font weight inside of the area type button. No need for everything to be bold.
- Removes extra margin above “Select an area type” heading.
- Removes extra margin below area type descriptions in the area type modal.

lucascumsille:
Thank you so much guys! I just rebased and merged it  =)

PR #531, opened 2024-04-17
Import and display action scorecards scores for local councils

### LINKED ISSUE #514: Import data on Local Authority Climate Action Scorecards
At the very least, it’d be good to have:

- a dataset for the council’s _total_ score as a % figure – which we can display as a big number on the council page
- a dataset for the council’s section scores (with a datatype per section) as % figures – which we can display as a table on the council page

We need to be careful that we _don’t_ display comparisons with a "national average" though. If possible, it’d be good to compare with the average of councils of the same type, but I’m not sure how we’d do this, given the way we currently handle metadata (min, max, average) on datasets.

struan:
Splits it into two datasets: total and section scores. 
Updates the area template to allow range related datasets.

Fixes #514

Screenshot 2024-04-17 at 14 39 24

### DIFF
--- a/hub/management/commands/import_council_scorecards_score.py
+++ b/hub/management/commands/import_council_scorecards_score.py
@@ -1,0 +1,167 @@
+from django.conf import settings
+
+import pandas as pd
+from tqdm import tqdm
+
+from hub.import_utils import filter_authority_type
+from hub.models import DataSet, DataType
+
+from .base_importers import BaseImportFromDataFrameCommand, MultipleAreaTypesMixin
+
+declare_map = {
+    "Y": "Yes",
+    "N": "No",
+}
+
+
+class Command(MultipleAreaTypesMixin, BaseImportFromDataFrameCommand):
+    cons_row = "gss_code"
+    message = "Importing council scorecards data"
+    uses_gss = True
+    do_not_convert = True
+
+    data_file = settings.BASE_DIR / "data" / "2023_scorecards_data.csv"
+
+    area_types = ["STC", "DIS"]
+
+    defaults = {
+        "label": "Council Climate Action Scorecard",
+        "description": "",
+        "data_type": "percent",
+        "category": "place",
+        "release_date": "2023",
+        "source_label": "Data from Climate Emergency UK.",
+        "source": "https://councilclimatescorecards.uk/",
+        "source_type": "csv",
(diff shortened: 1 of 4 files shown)


struan:
Not clear if the chart needs a title, possibly so.

zarino (changes requested):
This looks generally ok to me. Three points:

- Ideally, rather than comparing to a "UK average", we’d compare to all councils of the current council’s "type". But I expect there’s no way to easily do that right now. So, as a minimum, I think we should hide the "UK average" stuff _everywhere_ in relation to this data (so, on the area page card, and on the Explore page) – both because comparisons to the average mean less when different councils have different powers/responsibilities (and even different maximum possible scores per section) and we _also_ said to council officers, in the run-up to the Scorecards, that we wouldn’t facilitate comparisons outside a council’s "type".
- On the area page card, my instinct is a table (without the `js-chart` class) would be easier to read than a chart. What do you think?
- I’d title the card "Council Climate Action Scorecard" rather than "Action Scorecards overall score"

struan:
Tweaked:

Screenshot 2024-04-22 at 10 52 40

hub/management/commands/import_council_scorecards_score.py
```diff
@@ -0,0 +1,167 @@
...
+
+    area_types = ["STC", "DIS"]
+
+    defaults = {
+        "label": "Council Climate Action Scorecard",
+        "description": "",
+        "data_type": "percent",
+        "category": "place",
+        "release_date": "2023",
+        "source_label": "Data from the Climate Emergency UK.",
```
zarino, line 33: Drop the "the" 😉

hub/management/commands/import_council_scorecards_score.py
```diff
@@ -0,0 +1,167 @@
...
+        "source": "https://councilclimatescorecards.uk/",
```
zarino, line 34: I wonder, could we make this a direct link to the council’s page on the scorecards site? At least that way, there’ll be a direct link to more detail from the dataset card on the area page.

zarino (changes requested):
Two quick comments on things I’ve just noticed (sorry!)

I also wonder what’s going on with the rounding in your screenshot?

Screenshot 2024-04-22 at 10 58 24

PR #549, opened 2024-05-23
switch to using DC API for fetching 2024 PPCs

### LINKED ISSUE #548: change PPC import to use Democracy Club API instead of ad hoc spreadsheet
Now that we have a dare for the general election date the PPCs will be updated in the DC candidates app and not the spreadsheet so we need to update the importer to use this.

struan:
Fixes #548


zarino (approved):
Have tested this out locally, and it looks good to me! Nice to be able to include the latest data live from Democracy Club.

PR #554, opened 2024-06-14
PersonData party improvements

zarino:
An LIH user suggested we show MP and PPC parties on the search results page. Great idea!

While I was there, I realised getting hold of the party outside of the Area page was a bit of a pain, so I’ve suggested some code that adds `party` as a computed field on the `Person` model. Makes it just as easy to get a person’s `party` as it is to get their `name` or `area`. @struan – look ok to you?


struan (approved):
👍

PR #556, opened 2024-06-17
import member data from Sussex Wildlife Trust

### LINKED ISSUE #555: Import Sussex Wildlife Trust member totals
Sussex WT have provided their member totals per (new) constituency:

https://docs.google.com/spreadsheets/d/1gfhjo9z4o7AqiwG0ECByNofDgV8aow4uF4-PWyALmBk/edit

We’ll need to ask TCC to get a data sharing agreement in place with them, and to confirm whether the data should be public or members-only.

But in the meantime, can we get the data imported, ready?

I’m not sure whether it’s best to set this up as a sussex-specific dataset (ie: `sussex_wildlife_trust_members`) or just the first few rows of what we hope will eventually become a _national_ dataset (ie: `wildlife_trusts_members`).

struan:
Fixes #555

### DIFF
--- a/hub/management/commands/import_sussex_wt_member_count.py
+++ b/hub/management/commands/import_sussex_wt_member_count.py
@@ -1,0 +1,48 @@
+from django.conf import settings
+
+import pandas as pd
+
+from hub.models import DataSet
+
+from .base_importers import BaseImportFromDataFrameCommand
+
+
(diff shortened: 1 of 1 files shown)


zarino (approved):
# 👍 🎉 🦡

PR #563, opened 2024-06-21
Local postcode lookup

ajparsons:
Ports lookup tool from https://github.com/mysociety/mini-postcode-lookup/ and adds more explanatory text.

image

One thing I wasn't sure about was how to link it, I've gone for the approach of putting 'TOOLS' in the top menu which links to this for the moment. 

image


local_intelligence_hub/urls.py
```diff
@@ -57,6 +57,7 @@
         area.UnFavouriteDataSetView.as_view(),
         name="unfavourite_dataset",
     ),
+    path("tools/postcode", tools.PostcodeView.as_view(), name="postcode_converter"),
```
struan, line 60: everything else has a `/` at the end of the url so would be nice to have that for consistency.

hub/templates/hub/tools/postcode.html
```diff
@@ -0,0 +1,169 @@
...
+        document.getElementsByName("output_type").forEach(function (element) {
+            element.addEventListener("input", process);
+        });
+
+        document.getElementById("copyButton").addEventListener("click", function () {
+            var constituenciesText = document.getElementById("constituencies").value;
+            navigator.clipboard.writeText(constituenciesText)
+                .then(() => {
+                    console.log("Constituencies copied to clipboard");
+                })
```
struan, line 152: maybe remove this debug?

hub/templates/hub/tools/postcode.html
```diff
@@ -0,0 +1,169 @@
...
+            // change the copyButton class to btn-success and its text to 'Copied to clipboard'
+            // for two seconds before reverting to the original class and text
+            document.getElementById("copyButton").classList.remove("btn-primary");
+            document.getElementById("copyButton").classList.add("btn-success");
+            document.getElementById("copyButton").textContent = "Copied to clipboard";
+            setTimeout(function () {
+                document.getElementById("copyButton").classList.remove("btn-success");
+                document.getElementById("copyButton").classList.add("btn-primary");
+                document.getElementById("copyButton").textContent = "Copy Constituencies to clipboard";
+            }, 500);
```
struan, line 165: This does not seem like an obvious way to do this. I suspect having a message that says "copied to the clipboard" which is then removed if they add another postcode.

hub/views/tools.py
```diff
@@ -0,0 +1,14 @@
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import render
+from django.views import View
+
+
+class PostcodeView(View):
```
struan, line 6: I'd maybe just stick this in the core views file for now. we can split it out if we're going to be adding more tools.

hub/views/tools.py
```diff
@@ -0,0 +1,14 @@
...
+    def get(self, request: HttpRequest) -> HttpResponse:
```
struan, line 7: given that we're largely not using type hints elsewhere in the codebase I have a slight preference for not adding them to stray functions.

struan (approved):
A few tiny fussy things but otherwise, yay!

hub/templates/hub/tools/postcode.html
ajparsons: Yep, that's nicer

zarino:
@ajparsons has now implemented this separately, in Github Pages. We’ll link to it from a new "Tools" page on the Hub.

PR #565, opened 2024-06-24
import script for local council control

### LINKED ISSUE #202: Local authority council party control
[CAPE/Scorecards contains political control data](https://github.com/mysociety/caps/blob/5c6f6007292a5a06fd9db3127f6179a185433697/caps/management/commands/import_filter_data.py#L41) from January 2022 – [sourced from opencouncildata.co.uk](https://github.com/mysociety/climate_league_filter_datasets/blob/main/notebooks/political_control.ipynb). Could we update and re-use that?

struan:
Requires a .env entry for PARTY_CONTROL_URL

Only deals with Majority control although minority control is available.

Fixes #202

### DIFF
--- a/hub/management/commands/import_council_control.py
+++ b/hub/management/commands/import_council_control.py
@@ -1,0 +1,81 @@
+from datetime import date
+
+from django.conf import settings
+
+import pandas as pd
+
+from hub.import_utils import filter_authority_type
+from hub.models import DataSet
+
+from .base_importers import (
+    BaseImportFromDataFrameCommand,
+    MultipleAreaTypesMixin,
+    party_shades,
+)
+
+party_map = {
+    "CON": "Conservative Party",
+    "LAB": "Labout Party",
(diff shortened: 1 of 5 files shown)


hub/management/commands/import_council_control.py
```diff
@@ -0,0 +1,95 @@
...
+from django.conf import settings
+
+import pandas as pd
+
+from hub.import_utils import filter_authority_type
+from hub.models import DataSet
+
+from .base_importers import BaseImportFromDataFrameCommand, MultipleAreaTypesMixin
+
+party_shades = {
```
zarino, line 12: This’ll be the third import script that defines an identical `party_shades` dict. I wonder whether we should put into into a shared file (maybe even `.base_importers`?) that we can then import in each script that needs it?

zarino (approved):
Looks good. Just one suggestion about preventing duplication of `party_shades`.

PR #582, opened 2024-07-11
Add visible flag to datasets

struan:
Defaults to true, datasets isn't availble to filter, shade or on the area page if it's false.


zarino (changes requested):
Looks ok!

A few thoughts:

- You’re missing a `.filter(visible=True)` on the `DataSet.objects` in `views/core.py`’s `SourcesView` – if datasets aren’t visible, we probably don’t want to show them on the Sources page ;-)
- I wonder whether we want a `.filter(visible=True)` on the `DataSet.objects` in `views/core.py`’s `StatusView` too?
- Might be nice to add `visible` to the list of `list_filter`s in `admin.py`’s `DataSetAdmin`, so there’s an easy way to see just visible/invisible datasets in the Admin UI.

zarino (approved):
🎉

PR #583, opened 2024-07-16
Allow pinpointing of locations on Explore map

zarino:
Uses the Leaflet GeoSearch plugin, and OSM Nominatim geocoder, to allow users to place a marker on the map, at the requested location.

Jul-16-2024 08-37-31

Handy when you know the name of a place, but not which constituency it’s in!

Part of this work required me to have access to `this.map` inside a callback function, which revealed the same Vue.js proxying issue we’ve encountered in other projects. The fix is to ensure we “un-proxy” all references `this.map` with `toRaw(this.map)`.


hub/static/js/explore.esm.js
```diff
@@ -381,7 +383,7 @@ const app = createApp({
         // because the map element should be visible by then.
         var _this = this
         setTimeout(function(){
-          _this.map.invalidateSize()
+          toRaw(_this.map).invalidateSize()
```
struan, line 386: I'm not sure what the proxying issue is here so it might be nice to provide more details in either a comment even if it's just pointing somewhere else, or the commit message to explain what's going on?

zarino: Apols, yes, the proxying issue is that, when you store a `Leaflet.Map` instance in the Vue.js state (making it accessible inside your Vue methods as, say, `this.map`), Vue _actually_ stores a wrapped version of the original instance, to enable its clever reactive features. This “proxy” object, though, breaks some internal Leaflet stuff, including `animateZoom` (see https://stackoverflow.com/questions/65981712/uncaught-typeerror-this-map-is-null-vue-js-3-leaflet), which results in all sorts of weird bugs with zooming / panning / fitting to bounds. The recommendation from the Vue.js docs is to “un-wrap” that proxy object when you need to work with it, replacing `this.map` with `toRaw(this.map)`.

I didn’t _need_ to use `toRaw()` in every place that I’ve used it here (just inside the `geosearch/showlocation` callback would have done, because that’s the only place we zoom to bounds), but I figure it’s simpler to use it _everywhere_ we refer to `this.map` than to use it inconsitently and force future maintainers to keep making the decision as to whether it’s needed or not. Would you agree?

I’ll amend the commit message to include some/most of this! ☝️

PR #589, opened 2024-07-22
Fix election data loading and area import for Post 2023 GE

### LINKED ISSUE #579: post election fixes
- [x] Fix election results import so a person can have a result in an old and new constituency
- [x] Make sure persondata imports don't delete old constituency data when importing new constituency data
- [x] Update persondata imports areas available settings
- [ ] Fix duplicate MP detection
- [ ] Update election results script to point at correct HoC library source for 2024 GE once available
- [x] More permanent solution for getting correct person type than hardcode MP in the code

struan:
This updates some scripts with new data sources, as well as enabling person data to be linked to a specific area type which makes some display things a bit easer.

Note that at the moment the import_mps_election_results script doesn't work as the parliament API has not been updated.

Fixes #579


hub/tests/test_views.py
```diff
@@ -189,6 +246,17 @@ def test_explore_json_shader_custom_colours(self):
...
+    def test_explore_persondata_area_type(self):
+        url = reverse("explore_json")
+        response = self.client.get(url + "?mp_election_majority__gt=1000")
+        self.assertContains(response, "South Borsetshire")
+        self.assertNotContains(response, "New South Borsetshire")
+        response = self.client.get(
+            url + "?mp_election_majority__gt=1000&area_type=WMC23"
+        )
+        self.assertContains(response, "New South Borsetshire")
+        self.assertNotContains(response, '"South Borsetshire')
```
zarino, line 258: I assume the quoting here is intentional?

zarino (approved):
This looks good to me, although I haven’t run it locally. Just one question, inline, about some suspicious quote marks.

hub/tests/test_views.py
struan: Yup, because it's testing JSON it's checking for a JSON string that starts with `South Borsetshire` while not catching `New South Borsetshire`

PR #591, opened 2024-07-26
Temporarily hide MP positions and memberships card

zarino:
The data we have is out of date, but because these MP dataset cards are custom rendered, we can’t just hide the datasets in the Admin. So, this commit comments them out, until we have updated data.

Screenshot 2024-07-26 at 13 36 20


struan (approved):
👍

PR #594, opened 2024-07-29
fix PersonData filter that was causing person_area to be joined twice

### LINKED ISSUE #590: Explore table/csv view listing MP parties twice, for pre-2024 MPs
See screenshot, many MPs have two parties listed:

Screenshot 2024-07-25 at 13-45-23 Explore Local Intelligence Hub

struan:
Construct a single filter clause that handles all the filtering rather than use two filter clauses as the latter approach resulted in two join clauses in the SQL which was leading to repeated results in the table view.

Fixes #590


zarino (approved):
Have tested this out locally, and it fixes the issue! 🎉 👍

PR #606, opened 2024-09-17
don't return new modal dialogs if they're already defined

struan:
In theory I think vue should keep track of this but there's some sort of interaction that happens which means it can lose track of the dialog which means that e.g dialog.hide() does not work. To avoid this check for an existing dialog and return that if so then return that, otherwise create a new one.


zarino (approved):
I’m kind of amazed this works (the `datasetModal()` computed property just returning _erm_ itself?) but have tried it out and can confirm it does.

PR #607, opened 2024-09-17
Great Big Green Week 2024 import

### LINKED ISSUE #593: Import 2024 Great Big Green Week events
At the very least, import the count of 2024 GBGW events in each (2025) constituency.

If we’re able to, actually listing the _organisations_ / _groups_ that ran the events would be a nice counterpart to the Open Letter Signatory dataset.

When we import the data, TCC wants us to make it clear (in the dataset description) that "This dataset does not include private events."

struan:
Add a generic processing command to turn the spreadsheet of events into one with constituencies. Add a new command for the 2024 events.

Fixes #593

### DIFF
--- a/hub/management/commands/import_gbgw_events_24.py
+++ b/hub/management/commands/import_gbgw_events_24.py
@@ -1,0 +1,43 @@
+from django.conf import settings
+
+from hub.models import DataSet
+
+from .base_importers import BaseConstituencyCountImportCommand, MultipleAreaTypesMixin
+
+
+class Command(MultipleAreaTypesMixin, BaseConstituencyCountImportCommand):
+    help = "Import data about number of GBGW events in 2024 per constituency"
+    message = "Importing 2024 GBGW events"
+    uses_gss = False
+
+    data_file = settings.BASE_DIR / "data" / "gbgw_events_24_processed.csv"
+
+    area_types = ["WMC23", "STC", "DIS"]
+    cons_col_map = {
+        "WMC23": "WMC23",
+        "STC": "STC",
+        "DIS": "DIS",
+    }
+
+    data_sets = {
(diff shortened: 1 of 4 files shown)


hub/management/commands/import_gbgw_events_24.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class Command(MultipleAreaTypesMixin, BaseConstituencyCountImportCommand):
+    help = "Import data about number of GBGW events in 2024 per constituency"
+    message = "Importing 2024 GBGW events"
+    uses_gss = False
+
+    data_file = settings.BASE_DIR / "data" / "gbgw_events_24_processed.csv"
+
+    area_types = ["WMC", "WMC23", "STC", "DIS"]
```
zarino, line 15: When I generate `gbgw_events_24_processed.csv`, the `WMC` column is empty. I assume this is intentional, because geocoding a lonlat to pre-2024 constituencies with MapIt is a pain.

So I think you need to remove `"WMC"` from the `area_types` here, so that the base importer doesn’t try to create data for (pre-2024) WMCs.

If you leave `"WMC"` in this list, without a column of WMCs in `gbgw_events_24_processed.csv`, you’ll get this error:

```
  File "/app/hub/management/commands/base_importers.py", line 81, in get_cons_col
    return self.cons_col_map[self.area_type]
KeyError: 'WMC'
```

zarino (changes requested):
One bug, I think, and a couple of comments about whether there’s anything simple we can do to better document the format and necessity of the `--year` flag on `generate_gbgw_data.py`.

zarino (approved):
💥 BOOM! Love it. LGTM!

struan:
merged in 601b0cabf6b396c64c01e511f2a9e6e733892f3c

Issue #603, opened 2024-09-02
Investigate feasibility of importing Written Questions and Early Day Motions on climate/nature topics (from TheyWorkForYou?)


zarino:
Data on MPs voting on climate/nature bills is going to be thin on the ground for a while, and with so many new MPs this intake, we can’t rely on data from the previous parliament to give us a full picture either. So we need to looks for other indicators of MP support for / opposition of climate and nature issues, from their other activities in parliament.

Written Questions and Early Day Motions could be two such sources of data.

- Where do we get the data from?
- How do we filter down to just climate and nature related WQs/EDMs?
- How do we present it to be most useful?

zarino:
> Where do we get the data from?

Two options for Written Questions/Answers:

- [TheyWorkForYou](https://www.theyworkforyou.com/api/docs/getWrans)’s `/api/docs/getWrans` API endpoint
- [UK Parliament](https://questions-statements-api.parliament.uk/index.html)’s `/api/writtenquestions/questions` API endpoint

> How do we filter down to just climate and nature related WQ’s

My instinct would be to use API search params to return a combination of:

- all questions submitted to DESNZ, DEFRA, and maybe DfT?
- any questions submitted to other departments, which contain terms like `net zero`, `climate change`, `climate emergency`, `climate crisis`, `nature`, `biodiversity`, `pollution`, `decarbonisation`, `carbon capture`, `renewable energy`, `onshore wind`, `offshore wind`, `wind turbine`, `solar panel`, `solar farm`, etc

The former will catch most climate/nature related questions, while the latter will mop up any remainder submitted to the other departments, and hopefully, if we can make the search terms specific enough, we won’t get too many false-positives.

> How do we present it to be most useful?

Perhaps show the titles of the 3 most recent questions submitted by the constituency’s MP (each one linking to the full question on TheyWorkForYou or the UK Parliament site), and then a button that unfolds a full list in place?

ajparsons:
Related  https://github.com/mysociety/twfy-votes-django/issues/2

struan:
I think both of these are now done, at least in an initial form. Will probably need some improvements along the way.

The wrans this has a cron so should run every day and fetch the previous days wrans from TWFY. I might add a once a week to do the last seven days on a Sunday just as a backup.

PR #608, opened 2024-09-17
updated EDM import that uses a CSV as a source for relevant EDMs

struan:
Mostly the same code as the existing importer, only it uses a CSV to getg the list of EDMs to import so this can be updated without having to touch the code.

Part of #603

Screenshot 2024-09-17 at 14 22 55

### DIFF
--- a/hub/management/commands/import_mp_edm_signatures.py
+++ b/hub/management/commands/import_mp_edm_signatures.py
@@ -1,0 +1,170 @@
+from functools import cache
+
+from django.conf import settings
+from django.core.management.base import BaseCommand
+
+import pandas as pd
+import requests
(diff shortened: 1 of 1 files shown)


struan:
Example EDMs csv:

[relevant_edms.csv](https://github.com/user-attachments/files/17027736/relevant_edms.csv)

zarino (approved):
This works! (Once I remembered that I needed to mark the datasets as `is_public`, or log in, to see them.)

I wonder whether repeating "MP support for…" in the table on the area page is really required – although I can see why you’ve named the dataset that way. This is probably good enough to go for now, we can always rename the dataset, or whatever other improvement we come up with, later.

PR #609, opened 2024-09-17
fetch and save recent relevant MP written answers

struan:
This fetches yesterday's wrans XML from TheyWorkForYou and parses out relevant questions, defined by department, and updates this list for the MP to include the last three questions.

Screenshot 2024-09-18 at 12 10 32


zarino (approved):
The way we display this on the area page probably needs some tweaking (to avoid the three-column layout, for example, and to better explain what a Written Question is) but this is a good enough start!

PR #610, opened 2024-09-18
Added overflow hidden to dataset-card

lucascumsille:
Working on another ticket I noticed an odd horizontal overflowing when you are on mobile. After some playing around I realised that it only happens when there is a graph.

Fixes:

https://github.com/user-attachments/assets/dfdf2ffa-b493-4c41-8ea4-ede5ca51395b

### DIFF
--- a/hub/static/css/_dataset-card.scss
+++ b/hub/static/css/_dataset-card.scss
@@ -1,4 +1,7 @@
 .dataset-card {
+    // This prevents some odd overflow horizontal behaviour when there are graphs present on mobile.
+    overflow: hidden;
+
     .js-favourite-this-dataset {
         display: block;
     }


zarino:
Thanks for this @lucascumsille – looking closer, it turns out it wasn‘t the _charts_ themselves causing this overflow, but the theoretically "hidden" tables beneath them.

We were using Bootstrap’s `visually-hidden` mixin to hide the tables, without making them invisible to screenreaders. But it turns out the `width: 1px` in `visually-hidden` doesn’t work on a `display: table` element (makes sense, sorta – you can’t make a table narrower than the narrowest content of each of its cells).

I’ve [just commited a fix to that instead](https://github.com/mysociety/local-intelligence-hub/commit/b5d35a2f04315e7e85ab10d9827825462478e5a5), so the `overflow: hidden` isn’t required.

PR #611, opened 2024-09-19
import command for bbowt member counts

struan:
Imports Berkshire, Buckinghamshire and Oxfordshire Wildlife Trust member counts.


zarino (approved):
Looks good to me!

PR #612, opened 2024-09-19
import script that uses json config

struan:
This should enable add simple imports by adding to the JSON rather than writing a new script.

Also includes some examples from ONS and Commons Library plus a utility command to pivot CSVs to help with ONS imports.

### DIFF
--- a/conf/imports.json
+++ b/conf/imports.json
@@ -1,0 +1,446 @@
+[
+  {
+    "name":"constituency_age_distribution",
+    "label":"Constituency age distribution",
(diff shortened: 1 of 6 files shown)


struan:
Note that once the ONS data is fetched (one file per variable) you need to pivot the data to get it into one row per constituency so:
```
./manage.py pivot_csv --infile 2024_travel_to_work_method.csv --outfile 2024_travel_to_work_method_pivoted.csv --column "Method used to travel to workplace (12 categories)" --value_column Observation --index "Post-2019 Westminster Parliamentary constituencies Code"
```

PR #613, opened 2024-09-24
Various fixes for import scripts

zarino:
- Reinstates handling of 2010 WMCs in import scripts, and removes the no longer required `import_new_constituencies` command
- Fixes a common bug where a number of import scripts used `QuerySet.get()` instead of `QuerySet.filter()`, which meant they failed when run on a database that contained both `WMC` and `WMC23` AreaTypes.
- Marks a number of import scripts as either outdated or broken, in the `run_all_import_scripts` management command – something to follow up on later!!
- Turns `script/import-all-data` into a dumb wrapper around `./manage.py run_all_import_scripts`, to reduce duplication

The upshot of this is that—as long as you have all the required source data in the `/data` directory—a fresh `docker compose up` and then `script/dev-command ./manage.py run_all_import_scripts` successfully gets you a largely complete copy of the live site. (What’s missing, of course, is any scripts skipped in `run_all_import_scripts`, and any `featured` or `is_public` states on DataSets, which must be set manually via the Django admin.)

@struan I wasn’t sure whether to leave your debugging commits in, or remove them? Also, should we try to fix the five broken import scripts (the `ValueError` and `JSONDecodeError` ones now listed in `run_all_import_scripts`) as part of this PR, or is that something for a separate ticket?


struan (approved):
👍

Issue #621, opened 2024-11-18
Fix JSONDecodeError from import scripts calling Parlparse


zarino:
While implementing https://github.com/mysociety/local-intelligence-hub/pull/613 I noticed a couple of import scripts were failing with a JSONDecodeError, because the Parlparse JSON file could not be found, aparently…?

- `import_christian_aid_group_locations`
- `import_onward_polling_data`

Maybe I was just missing something obvious? If not, we should fix this.

PR #623, opened 2024-11-26
Fix some import bugs

### LINKED ISSUE #622: Fix ValueError from import scripts testing the truthiness of a Pandas dataframe
While implementing https://github.com/mysociety/local-intelligence-hub/pull/613 I noticed a couple of import scripts were failing because they attempt to check the truthiness of a Pandas dataframe…?

- `import_air_quality_data`
- `import_mps_select_committee_membership`
- `import_nt_property_locations`

Maybe this worked in previous versions of Pandas, but doesn’t work any more? I’m not familiar enough with Pandas or these scripts, to know how best to fix them.

struan:
Deal with some bugs in import scripts due to either changes in Pandas or moving of data sources.

Fixes #622 and #621

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ You could alternatively run commands individually (again, from inside the contai
     ./manage.py import_areas
     ./manage.py import_mps
 
+Finally, you will want to log in to `/admin` and make a selection of datasets “Public”, so they appear to logged-out users, on the site.
+
 ### Running the tests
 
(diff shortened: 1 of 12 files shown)


zarino:
This looks good. Guess the best way to test it, is to drop my Docker volumes, and run a `docker compose up` from scratch. 😬 Will give it a go when I have a quiet afternoon.

PR #628, opened 2025-01-10
Import Transition Network groups

### LINKED ISSUE #389: Import data on Transition Towns groups that cover each constituency
Transition Towns groups are listed here: https://transitiongroups.org/search-groups/?country=gb

The coverage area of each group is not explicitly defined, but we could take a guess from their name / location. Would probably require manual classification – if we assume 30 seconds per group, for 330 groups, that’s just under 3 hours’ work.

We could contact Mike at Transition Together (who we know via our NLCF Growing Great Ideas cohort) to see whether TT has any better data on coverage areas?

zarino:
Closes #389.

I would have loved to use our json config-based importer for this, but it doesn’t handle group imports yet, and modifying it to inherit from `BaseConstituencyGroupListImportCommand` rather than `BaseImportFromDataFrameCommand` was beyond the scope of this ticket.

To import the groups, run:

```
script/manage generate_transition_network_groups_csv
script/manage import_transition_network_groups
```

Or just let both commands be run automatically as part of:

```
script/manage run_all_import_scripts --generate
```

### DIFF
--- a/hub/management/commands/generate_transition_network_groups_csv.py
+++ b/hub/management/commands/generate_transition_network_groups_csv.py
@@ -1,0 +1,52 @@
+from django.conf import settings
+
+import pandas as pd
+import requests
+
+from .base_generators import BaseLatLonGeneratorCommand
+
+
+class Command(BaseLatLonGeneratorCommand):
+    help = (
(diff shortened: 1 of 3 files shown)


struan (approved):
looks good

PR #632, opened 2025-01-22
fix long label being truncated by chartjs

### LINKED ISSUE #436: 'Neither' is truncated to 'her' in 'Public opinion' dataset
Url: https://www.localintelligencehub.com/area/WMC/Tiverton%20and%20Honiton

Card on "Likelihood to take action such as voting for a particular candidate or lobbying an MP to tackle climate change and protect nature" has text ` her likely nor unlikely`. I guess it should be `Neither likely nor unlikely`.

image

struan:
Split long labels into an array of strings no longer than 30 characters which chartjs will then render as multiple lines.

Fixes #436 (again)

### DIFF
--- a/hub/static/js/area.js
+++ b/hub/static/js/area.js
@@ -184,7 +184,7 @@ var makeChart = function() {
             layout: {
                 padding: {
                     // Some extra padding for the data labels.
-                    right: 30
+                    right: 35
                 }
             },
             scales: {
@@ -196,6 +196,9 @@ var makeChart = function() {
                     }
                 },
                 [primaryAxis]: {
+                    ticks: {
+                        autoSkip: false
+                    },
                     grid: {
                         display: false
                     }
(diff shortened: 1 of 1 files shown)


zarino (approved):
Thanks @struan. I discovered that if you set `ticks: { autoSkip: false }` on the axis, Chart.js will render every label, even if it thinks they’re going to be too closely smooshed together, which avoids the "chart.js is hiding labels when you set the max_length to something small like 20" problem.

I’ve tried this out locally, with a `max_label_length` of 20, and it seems to produce pretty good results on all the charts in my local environment, at least. Even labels nearing 60 characters have been fine (broken onto 3 lines).

I’ve also bumped up the right padding on the charts to 35px, as I noticed text was still sometimes getting cut off. (I think _this_ cutting off happens because Chart.js calculates chart dimensions based on the labels being set in Arial, but then we apply a custom font, which results in slightly wider labels.)

struan:
I think 20 might be too short as you get lines smashing into each other vertically. I've tweaked it to 28 which is maybe a reasonable compromise.

Screenshot 2025-01-27 at 14 10 11

PR #647, opened 2025-05-19
Semi public area actions

struan:
Allow area actions to be only viewable if you have a link with the correct query string


hub/views/area.py
```diff
@@ -408,10 +408,16 @@ def get_context_data(self, **kwargs):
         ).select_related("action")
 
         if self.request.session.get("area_action"):
-            actions = actions.filter(
-                action__require_session=True,
-                action__passphrase=self.request.session["area_action"],
-            )
+            area_key = f"{self.request.session['area_action']}_area"
```
zarino, line 411: Minor thing, but I wonder whether it’s worth naming this `…_area_slug` so that it’s a bit clearer that the value will be a slug?

struan: not doing this now

zarino (approved):

PR #657, opened 2025-06-20
[WIP] Coastal constituency data import

### LINKED ISSUE #183: Coastal constituencies (touching the coast)
May be easiest to eyeball this and compile a list manually.

zarino:
TODO:

- Make this actually create a boolean DataSet/DataType (it currently creates a DataSet and DataType with null data_type attributes).
- Make import_from_config.py actually pass the “description” field from the JSON along to the final DataSet (it currently doesn’t).
- Add boolean comparator to the Django, Vue.js, and explore.html template (there currently isn’t one).

Here’s the source data: [`coastal_constituencies_2024.csv`](https://github.com/user-attachments/files/20836585/coastal_constituencies_2024.csv)

Fixes #183.


zarino (commented):
@struan This mostly looks good now, with your most recent changes. Two things I noticed:

- The `description` from the record in `imports.json` isn’t pulled through to the `DataSet` in the database
- You get an internal exception (`function max(boolean) does not exist`) when you try to shade by coastal constituency, because I think it’s assuming the shader data will be numeric

PR #659, opened 2025-06-23
FoE act now data

### LINKED ISSUE #652: Import ‘Act Now Change Forever’ constituency briefing PDFs from Friends of the Earth
Toby from FoE has emailed over (to @zarino) the briefings, and a mapping from filename to constituency PCON24 code.

struan:
Fixes #652


zarino (approved):
@struan this looks good to me, and I can confirm the data import config worked.

FYI I’ve changed the expected CSV filename from `tcc-filename-lookup.csv` to `foe-act-now-filename-lookup.csv` just so that it’s more obvious what it is in months/years to come.

I also made the `DataSet` public by default.

PR #663, opened 2025-07-02
'Act Now Change Forever' Top banner

lucascumsille:
Preview:

https://github.com/user-attachments/assets/64964649-7553-4308-8904-b2546d48bdbd

I agree that adding a link to the PDF on the current banner, will compete with all the information we already have there. I thought adding this sticky banner would be better; it's noticeable enough thanks to the navy background (I got the colours from the ANCF infographic), and it doesn't clash with the colours we currently use.

I thought of a second approach, to add a sticky button at the bottom of the page that takes you to the pink ACNF card, but then I wondered if the button would give enough context for people to really click on it, only to be taken to the card and then click again to download the PDF.

Please note: Currently `_act_now_bar.html` is being always included on the `area.html` I haven't done the work for the backend.

### DIFF
--- a/hub/static/css/_act-now-banner.scss
+++ b/hub/static/css/_act-now-banner.scss
@@ -1,6 +1,8 @@
 .act-now-banner {
     position: relative;
     background: #ffc9ef;
+    border-radius: $border-radius;
+    overflow: hidden;
 
     display: flex;
     flex-direction: column;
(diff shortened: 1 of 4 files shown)


lucascumsille:
Perfect thank you @zarino I just merged and currently deploying =)

PR #664, opened 2025-07-02
convert constituency MRP data to council data

### LINKED ISSUE #661: Expose more MRP polling data at Single Tier and District council level
We currently only have the 2022 MRP datasets available for councils. We should use our automatic area conversion at import time to make the more recent MRP datasets available at council level too.

struan:
Updates the DataType convertor to enable transforming Westminster constituency level data to local council data, and then uses that in a script to do that.

Fixes #661


ajparsons (approved):
You've got a linting error but all the steps on this look good. 

(In principle we could update the lookup table to have the gss codes for new constituencies now - but the fix is already in so)

PR #665, opened 2025-07-08
More in common MRP dataset

### LINKED ISSUE #650: Import More In Common’s April 2025 voter intention MRP data
https://www.moreincommon.org.uk/latest-insights/more-in-common-s-april-mrp/

struan:
As well as the dataset this updates the base importer and the `import_from_config` command to enable skipping countries in an individual data type on the way in so we can avoid SNP and Plaid appearing outside Scotland/Wales.

Fixes #650

### DIFF
--- a/conf/imports.json
+++ b/conf/imports.json
@@ -1713,5 +1713,150 @@
         "data_col":"is_coastal"
       }
     ]
+  },
+  {
+    "name":"more_in_common_winner_mrp",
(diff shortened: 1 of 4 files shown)


zarino:
FWIW, More In Common has literally just [released a July 2025 update to this data](https://www.moreincommon.org.uk/latest-insights/more-in-common-s-july-mrp/). I assume we could just run off that instead of the April data?

PR #675, opened 2025-08-08
Add import config for Feb 2024 ECIU Net Zero economy data

zarino:
Data from https://eciu.net/analysis/reports/2024/the-uks-net-zero-economy-2024, inspired by conversations with Uplift UK, and participants at the recent TCC/Green Alliance event.

The Gross Value Add figures are massively skewed by City of London (where the inclusion of financial services as part of the "green economy" means that the City’s financial value outshines literally everything else in the UK by orders of magnitude. Not sure how useful this particular dataset will be, at least as a UK-wide shader on the Explore page, or for the City of London page specifically.

This got me wondering whether there’s a better way to handle outliers like this (eg: [the source page](https://eciu.net/analysis/reports/2024/the-uks-net-zero-economy-2024) from ECIU ranks the constituencies into quintiles, collapsing the massive gulf between the City of London and all the other constituencies.

But that’s probably something we could revisit in future. I think this new dataset is probably useful enough to publish as-is.

### DIFF
--- a/conf/imports.json
+++ b/conf/imports.json
@@ -1858,5 +1858,72 @@
         "data_col":"Other"
       }
     ]
+  },
+  {
+    "name": "eciu_net_zero_economy",
+    "label": null,
+    "description": null,
+    "data_type": null,
+    "is_range": false,
+    "category": "place",
+    "subcategory": null,
+    "release_date": "February 2024",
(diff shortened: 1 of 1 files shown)


struan (changes requested):
This seems to be using the pre 2024 Scottish constituency names so it's missing a lot of data for Scotland. It might be an idea to just exclude Scotland because I'm not sure that re-mapping the data is going to be valid.

PR #734, opened 2026-08-12
Add mySociety Newsletter

lucascumsille:
Fixes part of: https://github.com/mysociety/Design-Internal-tasks/issues/60 

### Footer version 1
Screenshot 2026-08-12 at 09 06 51

### Footer Version 2
Screenshot 2026-08-12 at 10 56 44

### About page
Screenshot 2026-08-12 at 10 55 42

Screenshot 2026-08-12 at 10 58 32

### DIFF
--- a/hub/forms.py
+++ b/hub/forms.py
@@ -163,7 +163,7 @@ class InactiveCheckLoginForm(AuthenticationForm):
 
 class MailingListSignupForm(Form):
     email = EmailField(label="Email")
-    full_name = CharField()
+    full_name = CharField(required=False)
     mysoc_signup = BooleanField(
         required=False,
         label=mark_safe(
--- a/hub/static/css/_newsletter.scss
+++ b/hub/static/css/_newsletter.scss
@@ -2,6 +2,7 @@
     display: flex;
     flex-direction: column;
     gap: 0.5rem;
+    max-width: 20rem;
 
     @include media-breakpoint-up(sm) {
         flex-direction: row;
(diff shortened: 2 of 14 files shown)


zarino (approved):
I think option 2 for the footer works best – three distinct “areas” with distinct purposes, side-by-side 👍

zarino:
Oh, by the way @lucascumsille @MyfanwyNixon I‘ve just remembered that there are _already_ mailchimp signup forms on the Local Intelligence Hub, at the bottom of every Area page:

Screenshot 2026-08-12 at 11-48-36 North Herefordshire

These forms add the user to _both_ mySociety and TCC’s mailchimp lists (assigning them to the Climate interest group on the mySociety list as you’d expect).

So maybe we should use that existing form backend for these new signup boxes? Rather than just hard-coding fields in the HTML?

lucascumsille:
@zarino Thanks for pointing that one out, I didn't see them. I just added two fixups that use the form you were mentioning. Let me know if I need to tweak anything else.

MyfanwyNixon:
> Want to know when we add more data?

@zarino Do we ever tell people that there have been updates on LIH? Not recently, I don't think! 
I feel like it might be more truthful to say 'updates on mySociety's climate-related work' or something.

zarino:
> Do we ever tell people that there have been updates on LIH? Not recently, I don't think!

We have, once or twice. But not recently, no. Partly because new data has been coming in dribs and drabs, never particularly newsworthy!

zarino (approved):
@lucascumsille I _think_ I’ve fixed the backend part of this so that it works. I also removed the duplication of all the different forms, so now it’s the same form everywhere, and AJAX submission should work across all of them too.

I’ve tested all of the forms manually (with the actual `Mailchimp.Client()` temporarily replaced with a no-op) and the forms all work. So I think you’re good to squash and merge this.

PR #735, opened 2026-09-02
Update to Django 5.2

struan:
Note the instructions wrt to updating the database in the first commit

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN curl -sSL https://install.python-poetry.org | python -
 ENV PATH="/root/.local/bin:$PATH"
 WORKDIR $PYSETUP_PATH
 COPY poetry.lock pyproject.toml ./
+RUN poetry config keyring.enabled false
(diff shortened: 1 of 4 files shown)


zarino (commented):
Weirdly, when I run this on my local install (including all of the steps in the postgres 15 commit message) none of the areas or datasets in my database seem to work (no areas are returned when searching, `Area not found` when loading an area page by URL, and error about `name_plural` on `/explore`). Whereas they all work if I revert back to `main`, rebuild the `web` image, and restore the SQL dump. 🤔 

There are a number of errors when I load the database dump, maybe that’s related? eg:

```
ERROR:  relation "auth_group_name_key" already exists
ERROR:  relation "auth_group_permissions_group_id_permission_id_0cd325b0_uniq" already exists
ERROR:  multiple primary keys for table "auth_group_permissions" are not allowed
ERROR:  multiple primary keys for table "auth_group" are not allowed
…
```

PS, `postgress` typo in the postgres 15 commit message.

